PR #5, opened 2024-10-14
feat: Add Kubeflow + MLFlow Terraform solution

### LINKED ISSUE #1056: Create a full terraform solution for `kubeflow`+`mlflow`
### Context

To align with the terraform modules organization spec, we need to provide a terraform solution for Charmed Kubeflow + Charmed MLFlow

### What needs to get done

Create a full terraform solution that complies with the [spec](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit?usp=drivesdk) using all the terraform modules created for the charms.

### Definition of Done

full terraform solution is created

orfeas-k:
This PR adds a Terraform solution module for Charmed Kubeflow + MLflow. Please refer to the README.md file for information on inputs, outputs and instructions on how this works.

### Testing
I tested deploying CKF + MLFlow using the solution and ran all the UATs (kf+mlflow+e2e) successfully. Feel free to test  following instructions in the README. I used the following command:
```
terraform apply -var cos_configuration=true -var dex_static_username="admin" -var dex_static_password="admin" -var kfp_db_revision=203 -var katib_db_revision=203 -var mlflow_mysql_revision=203  
```
Note that:
* due to https://github.com/canonical/mysql-k8s-operator/issues/504, we need to deploy using a pinned version of mysql-k8s (the one from `8.0/beta`) so we 'd need to run the following in order to deploy with COS configuration enabled.
* due to adding, dex username and password in https://github.com/canonical/charmed-kubeflow-solutions/pull/6, this should be merged in order to test this PR. Otherwise, we need to modify the module `source` to point to that PR's branch. 

Closes canonical/bundle-kubeflow#1056


modules/kubeflow-mlflow/outputs.tf
misohu, line 1: I think we should add to output reuires and providers for resource dispatcher .... because we may want to integratewith ngc charm like here https://charmed-kubeflow.io/docs/launch-ngc-notebooks. BTW we dont need the ngc as an option in this terraform ?

orfeas-k: Good catch, pushed a commit for resource dispatcher. Re ngc, I asked the team so let's see what we conclude on that.

orfeas-k: We decided that this is an optional integration, so we shouldn't include it in the solution

misohu (approved):
LGTM my deployment worked no problems

PR #2, opened 2024-10-08
feat: Add kubeflow Terraform solution

orfeas-k:
This PR adds a Terraform solution module for Charmed Kubeflow bundle. Please refer to the README.md file for information on inputs, outputs and instructions on how this works.

### History
This replaces https://github.com/canonical/analytics-terraform-solutions/pull/2. For a full history of commits, as well as previous reviews, please refer to that PR.

### Testing
I tested deploying CKF using the solution and ran the UATs successfully. Feel free to test on your own by following instructions in the README.

### COS integration
Here's the [integrate CKF with COS guide](https://charmed-kubeflow.io/docs/integrate-with-cos)  for reference. Note that due to https://github.com/canonical/mysql-k8s-operator/issues/504, we need to deploy using a pinned version of mysql-k8s (the one from `8.0/beta`) so we 'd need to run the following in order to deploy with COS configuration enabled.
```
terraform apply -var cos_configuration=true -var kfp_db_revision=203 -var katib_db_revision=203
```

Closes canonical/bundle-kubeflow/issues/1073

PR #6, opened 2024-10-15
feat(kubeflow): Enable dex-auth static username and password

orfeas-k:
Follow up PR to #2, enabling static username and password variables for Kubeflow solution module. This will omit the need for configuration afterwards (ie [this step](https://charmed-kubeflow.io/docs/install#set-credentials-for-a-static-user)) and enable deployment of the solution like:
```
terraform apply -var dex_static_username="admin" -var dex_static_password="admin"
```

Note that setting those variables to [sensitive](https://developer.hashicorp.com/terraform/language/values/variables#suppressing-values-in-cli-output) hides their values from the CLI print, although they will still be available at the Terraform state file. 
```
  + resource "juju_application" "dex_auth" {
      + config      = {
          + "connectors"      = ""
          + "public-url"      = "http://dex-auth.kubeflow.svc:5556"
          + "static-password" = (sensitive value)
          + "static-username" = (sensitive value)
        }
```
Ref canonical/bundle-kubeflow#1073


misohu (approved):
Looking good :)

PR #7, opened 2024-10-16
feat: Add Deploy to AKS scheduled workflow

### LINKED ISSUE #3: ci: Spec and implement workflow that applies Terraform solution
### Context

We need to implement CI that confirms the solution's functionality. Reaching out to SQA as to what should be checked before passing our solution to them, we captured their feedback below:

### Assertions
* Assert deployment is active and run UATs
* (stretch) Perform a clean at the end of the deployment. This is not a blocker since we think there are some `juju` issues for cleaning but it can provide some valuable data about hooks running.

Regarding this, SQA team is working on a spec that defines these expectations and a universal way to provide tests for a Terraform solution so long-term the above can change. More specific tests could be provided in the future, like breaking relations, or removing units, but those are not available yet.

### Frequency
* Weekly running the tests and before updating charms from beta/candidate to `stable`

### What needs to get done

Spec and create a workflow according to description above. Specing out can be done as part of this issue since this will be really similar to what we already have in bundle-kubeflow [deploy-to-eks](https://github.com/canonical/bundle-kubeflow/blob/main/.github/workflows/deploy-to-eks.yaml) or [deploy-to-aks](https://github.com/canonical/bundle-kubeflow/blob/main/.github/workflows/deploy-to-aks.yaml).  

### Definition of Done

There is workflow that checks the solution's functionality

orfeas-k:
Add workflow to:
* deploy Charmed Kubeflow + MLflow Terraform solution module 
  to an AKS cluster on workflow and scheduled
* assert the deployment goes to active (except grafana-agent-k8s)
* run all the UATs
 
This workflow has been defined in this part of this [solution spec](https://docs.google.com/document/d/19lJygvpy4GVqAojVMlG661me8pcrEbP5L2gXbVUvLuE/edit?tab=t.0#heading=h.8uw5vi413l5k). It has been based on what we do in bundle-kubeflow [deploy-to-aks.yaml](https://github.com/canonical/bundle-kubeflow/blob/main/.github/workflows/deploy-to-aks.yaml) and mlflow [deploy-eks.yaml](https://github.com/canonical/mlflow-operator/blob/main/.github/workflows/deploy-eks.yaml),

Successful run: https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/11403914192/job/31731957275?pr=7#step:13:241

Closes #3


modules/kubeflow-mlflow/tests/test_deployment.py
```diff
@@ -0,0 +1,131 @@
...
+                "mlflow-mysql",
+                "mlflow-server",
+                "mlmd",
+                "oidc-gatekeeper",
+                "pvcviewer-operator",
+                "resource-dispatcher",
+                "tensorboard-controller",
+                "tensorboards-web-app",
+                "training-operator",
+            ],
```
NohaIhab: suggestion: can we get the list of deployed charms dynamically and remove `grafana-agent-k8s` from there? It's likely that we miss something in this list if any changes were to occur to the deployed charms

.github/workflows/deploy-to-aks.yaml
```diff
@@ -0,0 +1,157 @@
...
+          sudo apt update -y
+          sudo apt install python${{ env.PYTHON_VERSION }} python${{ env.PYTHON_VERSION }}-distutils python${{ env.PYTHON_VERSION }}-venv -y
+
+      - name: Install CLI tools
+        run: |
+          wget https://bootstrap.pypa.io/get-pip.py
+          python${{ env.PYTHON_VERSION }} get-pip.py
+          python${{ env.PYTHON_VERSION }} -m pip install tox
+          sudo snap install juju --classic --channel=${{ env.JUJU_VERSION }}/stable
+          sudo snap install charmcraft --classic
```
NohaIhab, line 52: let's check if this is needed

.github/workflows/deploy-to-aks.yaml
```diff
@@ -0,0 +1,157 @@
...
+          sudo snap install terraform --classic
```
NohaIhab: better to pin the snap version

.github/workflows/deploy-to-aks.yaml
```diff
@@ -0,0 +1,157 @@
...
+          sudo add-apt-repository ppa:deadsnakes/ppa -y
+          sudo apt update -y
+          sudo apt install python${{ env.PYTHON_VERSION }} python${{ env.PYTHON_VERSION }}-distutils python${{ env.PYTHON_VERSION }}-venv -y
+
+      - name: Install CLI tools
+        run: |
+          wget https://bootstrap.pypa.io/get-pip.py
+          python${{ env.PYTHON_VERSION }} get-pip.py
+          python${{ env.PYTHON_VERSION }} -m pip install tox
+          sudo snap install juju --classic --channel=${{ env.JUJU_VERSION }}/stable
```
NohaIhab: i think juju 3.x snap is not classic confinement

.github/workflows/deploy-to-aks.yaml
```diff
@@ -0,0 +1,157 @@
...
+          file_path: ".github/dependencies.yaml"
+
+      - name: Update ENV variables from inputs if available
+        run: |
+          K8S_VERSION=${{ inputs.k8s_version || env.K8S_VERSION }}
+          echo "K8S_VERSION=${K8S_VERSION}" >> $GITHUB_ENV
+          UATS_BRANCH=${{ inputs.uats_branch || env.UATS_BRANCH }}
+          echo "UATS_BRANCH=${UATS_BRANCH}" >> $GITHUB_ENV
+
+      # Remove once https://github.com/canonical/bundle-kubeflow/issues/761
```
NohaIhab: does it make more sense to link https://github.com/canonical/charmed-kubeflow-uats/issues/79 here?

NohaIhab (commented):
great work @orfeas-k 
a few comments

modules/kubeflow-mlflow/tests/test_deployment.py
orfeas-k: Very good idea actually, updated!

.github/workflows/deploy-to-aks.yaml
orfeas-k: Pinned terraform to latest/stable

.github/workflows/deploy-to-aks.yaml
orfeas-k: Fixed

.github/workflows/deploy-to-aks.yaml
orfeas-k: Turns out that this is needed for ops_test https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/11403458661/job/31730574559?pr=7#step:12:3946

NohaIhab (approved):
thanks @orfeas-k !

PR #13, opened 2025-01-30
fix(ckf): Add outputs dashboard_links_provider and ingress_provider

orfeas-k:
Add outputs dashboard_links_provider and ingress_provider in order to enable integrating MLflow with kubeflow-dashboard and adding its link on the dashboard's left sidebar.

Ref canonical/charmed-mlflow-solutions#6


misohu (approved):
Changes looking good, we agreed with @orfeas-k he will first run the full deployment before merging.

PR #17, opened 2025-02-14
feat: providing model name as input

deusebio:


dparv:
I've tested this and it works as expected.

orfeas-k (commented):
Introducing a feature for creating the model or not option sounds good to me. However,
`kubeflow` solution doesn't have an input right now for model name. It rather has a constant model (`local.model`) that is hardcoded to `kubeflow`. The issue with the current implementation is that if the juju model is created by terraform and `kubeflow-mlflow` solution doesn't depend on it (in a way that Terraform is aware of), then there's the possibility that it will try to apply things to `kubeflow` model, before the model exists. Thus, I suggest that we introduce a `model` variable/input both to `kubeflow-mlflow` solution and to `kubeflow` solution. In the latter case, that 'd replace the [local.model we have now](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.9/modules/kubeflow/main.tf#L6-L8). If you need guidance, you can take a look at how we 're doing it in [mlflow-solution](https://github.com/canonical/charmed-mlflow-solutions/tree/track/2.15/modules/mlflow). We just need to be cautious to handle `create_model` and `model` right. The only downside of this is that we enable a `model` input which allows users to deploy to a model of a name different than kubeflow, which would break the solution. Thus, we 'd need to outline this in the README. Lastly, let's make sure to update the README accordingly depending on the changes we do.

orfeas-k (approved):

deusebio:
me and @orfeas-k discussed this privately. The point that @orfeas-k raises is sensible, but we have two options:

1. Enforcing dependency by introducing a variable for the model name, but Kubeflow requires the namespace to be called "kubeflow", and this cannot be enforced. 
2. Enforcing the model name (and therefore not use the variable), but this does not prevent the user to "forget" the dependency with `depends_on`. 

Between the two, we would rather prefer 2 (which is the current implementation), since despite both would produce problems (for 1 the deployment would not be functional, while 2 the deployment would fail), but because it is more confusing to have a variable that cannot be set, than forgetting a dependency specification which is logically reasonable (if the model is not created, it means that it needs to be created before, and this module depends on it). Beside, the feedback loop of 2 would actually be faster, as the deployments would fail very soon.

PR #22, opened 2025-03-10
[KF-7101] chore: update kfp version to track/2.3

### LINKED ISSUE #21: track/1.9 is not up-to-date with the 1.9 bundle
### Bug Description

When we deploy Kubeflow using the terraform plan in this repo in branch `track/1.9`, Charmed Kubeflow UATs are currently failing in Solutions QA at `test_bundle_correctness`.

I believe the issue arises because these terraform plans have not been updated according to the latest Charmed-Kubeflow bundle.

### To Reproduce

1. Deploy the terraform solution in AKS
2. Clone and run the UATs from track/1.9 (with `charmcraft` snap installed)

### Environment

> Let me know if you need this.

### Relevant Log Output

```shell
=========================== short test summary info ============================
FAILED driver/test_kubeflow_workloads.py::test_bundle_correctness - AssertionError: Failed bundle correctness check. Expected: 2.3/* Found: 2.2/stable
assert None
 +  where None = <built-in method match of re.Pattern object at 0x7f393af60450>('2.2/stable')
 +    where <built-in method match of re.Pattern object at 0x7f393af60450> = re.compile('2.3/*').match
 +      where re.compile('2.3/*') = <function compile at 0x7f393f895ea0>('2.3/*')
 +        where <function compile at 0x7f393f895ea0> = re.compile
============= 1 failed, 1 skipped, 1 xfailed in 6003.42s (1:40:03) =============
```

### Additional Context

_No response_

deusebio:
fixes #21

### DIFF
--- a/modules/kubeflow/applications.tf
+++ b/modules/kubeflow/applications.tf
@@ -95,7 +95,7 @@ module "katib_ui" {
 }
 
 module "kfp_api" {
-  source     = "git::https://github.com/canonical/kfp-operators//charms/kfp-api/terraform?ref=kf-7159-set-channel"
+  source     = "git::https://github.com/canonical/kfp-operators//charms/kfp-api/terraform?ref=track/2.3"
   model_name = var.create_model ? juju_model.kubeflow[0].name : local.model
   revision   = var.kfp_api_revision
 }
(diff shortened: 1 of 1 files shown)


deusebio:
Before merging this, we must also make sure we also backport the PR for updating the terraform provider on the kfp track 2.3, see https://github.com/canonical/kfp-operators/pull/683

Otherwise the solution tf provider version may rollback to 0.14.0 again

orfeas-k (dismissed):
Approving. Let's make sure we let people using the solution be aware of the bump (or test it before we merge as you already mentioned). Also, this https://github.com/canonical/charmed-kubeflow-solutions/issues/21#issuecomment-2714616723 mentions that the issue wasn't fixed.

deusebio:
@orfeas-k please re-review. The issue here is that we were not pinning the channel, and therefore we are picking the default channel from charmhub (which is 2.2 :O) 

I really believe we should pin channels either here or on the single charm tf modules (also for all other components beside kfp), otherwise what we deploy here is rather unstable: changing the default channel on charmhub would pull the carpet in the tf module that all of a sudden would be deploying something different. 

To keep things confined in this PR, I would just upgrade the KFP and put the channel for those. I would honestly address adding channel for the other charms in a separate PR

orfeas-k:
> The issue here is that we were not pinning the channel, and therefore we are picking the default channel from charmhub  

This caught me completely off-guard, as this should never be the default behavior. In the past, we had agreed to version modules using branches, which means that terraform module in `track/x` branch is guaranteed to be compatible (and thus meant to be used) only with charm from `x/stable` channel (see e.g. this [issue](https://github.com/canonical/kubeflow-volumes-operator/issues/159) and the linked [backport PR](https://github.com/canonical/kubeflow-volumes-operator/pull/161/files#diff-b25e8cb262b29f5005a96d0c4a06a3deb77698fa6e11e4f43b21f6f4ad0f45e4R10)). This is also implicitly visible in our [terraform module update policy spec](https://docs.google.com/document/d/1zG9uG_ww6OXGYnKxHibtvwLoAMzCVqLAB7eDKoGPedQ/edit?tab=t.0#heading=h.my0h52xzfw7l). 

I went through all our charm repos and their track branches and it looks like we 've missed setting the `channel` **only in kfp** (see e.g. [knative](https://github.com/canonical/knative-operators/blob/track/1.12/charms/knative-serving/terraform/variables.tf#L10)). Thus, instead of setting the `channel` in the solution module, I suggest we just fix the `track/2.3` (and maybe 2.2 too?) branch in kfp. Filed https://github.com/canonical/kfp-operators/issues/685

deusebio:
@orfeas-k I have rolled back the channel pinning, and put the pinning in the kfp upstream component, see https://github.com/canonical/kfp-operators/pull/688

I have also created the commit [9866ee8](https://github.com/canonical/charmed-kubeflow-solutions/pull/22/commits/9866ee87fd427e9cfbc4c1f8857ab7aa141ad043) to pin to that dev branch, in order to be able to test. That commit will be then rolled back once the PR is merged

orfeas-k:
@deusebio this looks good, ping me to approve once this is tested and pointing to 2.3.

deusebio:
Diko just gave me his +1 to merge this. Could someone else approve since @orfeas-k is off this week?

DnPlas (approved):

PR #25, opened 2025-03-28
[MISC] Risk handling for solutions

deusebio:
The PR proposes how to handle risk in the product bundle. The idea here is that:

1. In the bundle solution we don't want to support customizing tracks, as the track is decided by the solution itself (i.e. `kubeflow` 1.9 does not support ANY track for `kubeflow-pipeline`, but just the one we provide)
2. Allow customization of the risk (which gets propagated throughout channels in the bundle to make sure we are consistent, such that an edge deployment means all components on the edge track, same for beta, candidate and edge)
3. Allow still pinning of single revisions for charms, such that the product bundle can be used in deployments in a reproducible and consistent way

Initially I tried to have the high-level versions for all components (kfp, katib, etc) in a single file as locals variable, like

```
locals {
    track = {
        kfp = "2.3"
        katib = "0.17"
        ...
   }
}
```

To then use this in the channel and source, like:

```
module "katib_db_manager" {
  source     = "git::https://github.com/canonical/katib-operators//charms/katib-db-manager/terraform?ref=track/${locals.track.katib}"
  revision   = var.katib_db_manager_revision
  channel    = "${locals.track.katib}/${var.risk}"
}
```

However, `source` must be [hardcoded in terraform](https://github.com/hashicorp/terraform/issues/1439), but [seems to be possible in opentofu](https://github.com/opentofu/opentofu/issues/1042). Given this, I don't think it is worth using the locals for two reasons:
1. Channel updates still requires to update the track in the source
2. Moving the definition away from the module may increase the risk of inconsistencies (updating locals but not the source). Having them very close to each other will allow to spot bugs and inconsistencies a lot more IMHO

### DIFF
--- a/modules/kubeflow-mlflow/main.tf
+++ b/modules/kubeflow-mlflow/main.tf
@@ -59,7 +59,7 @@ module "kubeflow" {
 }
 
 module "mlflow" {
-  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=wip-backports"
+  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=track/2.15"
   # kubeflow module creates the model
   create_model                = false
   model                       = module.kubeflow.model
--- a/modules/kubeflow-mlflow/main.tf
+++ b/modules/kubeflow-mlflow/main.tf
@@ -1,5 +1,6 @@
 module "kubeflow" {
   source                           = "../kubeflow"
+  risk                             = var.risk
   create_model                     = var.create_model
   cos_configuration                = var.cos_configuration
   dex_connectors                   = var.dex_connectors
@@ -77,4 +78,5 @@ module "resource_dispatcher" {
   source     = "git::https://github.com/canonical/resource-dispatcher//terraform?ref=track/2.0"
   model_name = module.kubeflow.model
   revision   = var.resource_dispatcher_revision
+  channel    = "2.0/${var.risk}"
 }
--- a/modules/kubeflow-mlflow/main.tf
+++ b/modules/kubeflow-mlflow/main.tf
@@ -59,7 +59,7 @@ module "kubeflow" {
 }
 
 module "mlflow" {
-  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=wip-new-pinning"
+  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=wip-backports"
   # kubeflow module creates the model
   create_model                = false
   model                       = module.kubeflow.model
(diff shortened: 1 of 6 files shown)


ghislainbourgeois:
> ```
> module "katib_db_manager" {
>   source     = "git::https://github.com/canonical/katib-operators//charms/katib-db-manager/terraform?ref=track/${locals.track.katib}"
>   model_name = var.create_model ? juju_model.kubeflow[0].name : local.model
>   revision   = var.katib_db_manager_revision
>   channel    = "${locals.track.katib}/${var.risk}"
> }
> ```
> 
> However, `source` must be [hardcoded in terraform](https://github.com/hashicorp/terraform/issues/1439), but [seems to be possible in opentofu](https://github.com/opentofu/opentofu/issues/1042).

Regarding this particular issue, the suggested move to a proper Terraform registry will help, as modules in registries can be versioned in a way that Terraform understands, and the version can then be controlled by variables.

Thanks for this PR, I will take a full look later today.

deusebio:
@ghislainbourgeois Thank you for the comment. That's indeed one way to go. Are you currently using TF registry?

ghislainbourgeois:
> @ghislainbourgeois Thank you for the comment. That's indeed one way to go. Are you currently using TF registry?

We are not yet using a TF registry, but we are evaluating one in the working group.

modules/kubeflow/variables.tf
```diff
@@ -1,3 +1,14 @@
+variable "risk" {
+  type        = string
+  description = "Value for the risk to be used"
+  default     = "stable"
+
+  validation {
+    condition     = contains(["stable", "candidate", "beta", "edge"], var.risk)
+    error_message = "Valid values for var: risk are (stable, candidate, beta and edge)."
+  }
```
orfeas-k, line 9: kudos! we could probably validate more inputs.

orfeas-k:
I like the rationale of tihs, nothing to add. The changes LGTM, I 'm not approving since this is in wip state.

deusebio:
backported PR to #54

modules/kubeflow-mlflow/main.tf
```diff
@@ -58,10 +59,11 @@ module "kubeflow" {
 }
 
 module "mlflow" {
-  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=track/2.15"
+  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=wip-new-pinning"
```
deusebio: **todo** We should make sure that we a proper branch. But this requires to also backport the risk changes to track/2.15 on mlflow solution. 

Should we do that?

deusebio: Raised companion PR on : https://github.com/canonical/charmed-mlflow-solutions/pull/17

deusebio: Merged the PR and update the reference here!

NohaIhab (approved):

PR #33, opened 2025-04-28
[KF-7085] fix: kubeflow-mlflow deploys two grafana-agent instances

deusebio:
Fixes https://github.com/canonical/charmed-kubeflow-solutions/issues/20

However, this has also required to land a small modification in the charmed-kubeflow UATs:

https://github.com/canonical/charmed-kubeflow-uats/pull/177

The UATs has successfully passed in this run (where the new dev branch in the UATs was used): https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/14782452380/job/41504153376

PR #34, opened 2025-05-02
[KF-7085] fix: kubeflow-mlflow deploys two grafana-agent instances (#33)

deusebio:
backporting #33


orfeas-k (approved):
Let's merge using a merge commit!

PR #35, opened 2025-05-28
feat: add Kubeflow + Feast Terraform solution

### LINKED ISSUE #14: Feast bundle
### Context

Write a feast bundle to deploy feast with its dependencies. Details for the bundle charms can be found [here](https://docs.google.com/document/d/1rHF7nixCq5U9NW6OsN8OAjkQt2O2NUFLACuM0QOHGRw/edit?tab=t.0#heading=h.emzc47ymtmgm) 

### What needs to get done

1. Write a feast terraform module for bundle
2. Release the bundle to charmhub
3. Write bundle integration tests

### Definition of Done

1. Feast terraform module for bundle is written
2. Bundle is released to charmhub
3. Bundle integration tests are passing

NohaIhab:
Closes https://github.com/canonical/feast-operators/issues/14

Creates a terraform solution for deploying Charmed Kubeflow integrated with Feast

## Testing
1. Setup microk8s and juju, [follow the tutorial](https://charmed-kubeflow.io/docs/tutorial) for instructions
2. Clone the repo and checkout the PR branch
```
git clone  https://github.com/canonical/charmed-kubeflow-solutions.git
cd charmed-kubeflow-solutions
git checkout kf-7402-add-kubeflow-feast-solution
```
3. Install terraform snap
```
snap install terraform --classic
```
4. Initialize terraform
```
cd modules/kubeflow-feast
terraform init
```
5. Apply terraform solution
```
terraform apply -var cos_configuration=true -auto-approve
```


modules/kubeflow-feast/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Usage
+
+This solution module is intended to be used either on its own or as part of a higher-level module. 
+
+### Model
+This solution always creates a model of the name `kubeflow`, since Charmed Kubeflow cannot be deployed in a different model.
+
+### COS configuration
+
```
orfeas-k, line 54: I would also include a note about existing grafana agent input, similar to the following. My guess is that `kubeflow-mlflow` does not include it because initially it didn't exist but was introduced later per request. https://github.com/canonical/charmed-kubeflow-solutions/blob/5c8a6d4237701953ce3765c9d39fb1736be6e3c0/modules/kubeflow/README.md#L63-L70.

modules/kubeflow-feast/applications.tf
```diff
@@ -0,0 +1,61 @@
+module "feast_integrator" {
+  # tflint-ignore: terraform_module_pinned_source
+  source     = "git::https://github.com/canonical/feast-operators//charms/feast-integrator/terraform?ref=main"
+  model_name = module.kubeflow.model
+  revision   = var.feast_integrator_revision
```
orfeas-k, line 5: What about channels for feast charms? We had standardized in `kubeflow` solution to include those in versioned modules to ensure that the module isn't deploying something different. I can see that it's not needed since we can also rely on the branch, but yeah.

modules/kubeflow-feast/applications.tf
```diff
@@ -0,0 +1,61 @@
...
+  model_name = module.kubeflow.model
+  revision   = var.feast_ui_revision
+}
+
+module "offline_store" {
+  # tflint-ignore: terraform_module_pinned_source
+  source          = "git::https://github.com/canonical/postgresql-k8s-operator//terraform?ref=main"
+  juju_model_name = module.kubeflow.model
+  app_name        = "offline-store"
+  channel         = "14/stable"
```
orfeas-k: suggestion(non-blocking): what about prefixing with `feast-`? In order to make their role more obvious. Still, I 'm a bit ignorant to how feast integrates with Kubeflow so if it makes sense to keep it that way, SGTM.

modules/kubeflow-feast/integrations.tf
```diff
@@ -0,0 +1,111 @@
+resource "juju_integration" "feast_integrator_offline_store" {
```
orfeas-k: General question, feast charms do not provide any cos integration, right?

orfeas-k (commented):
Left some comments, nothing worrying though. Deployed without an issue, awesome work :)

modules/kubeflow-feast/integrations.tf
NohaIhab: no, it's not planned for this cycle

modules/kubeflow-feast/applications.tf
NohaIhab: good point

modules/kubeflow-feast/README.md
```diff
@@ -1,7 +1,7 @@
 # Charmed Kubeflow + Feast Terraform solution
 
 
-This is a Terraform module facilitating the deployment and integration of Charmed Kubeflow and Feast, using the [Terraform juju provider](https://github.com/juju/terraform-provider-juju/). For more information, refer to the provider [documentation](https://registry.terraform.io/providers/juju/juju/latest/docs). 
+This is a Terraform module facilitating the deployment and integration of Charmed Kubeflow and Feast, using the [Terraform juju provider](https://github.com/juju/terraform-provider-juju/). For more information, refer to the provider [documentation](https://feast_registry.terraform.io/providers/juju/juju/latest/docs).
```
orfeas-k:
```suggestion
This is a Terraform module facilitating the deployment and integration of Charmed Kubeflow and Feast, using the [Terraform juju provider](https://github.com/juju/terraform-provider-juju/). For more information, refer to the provider [documentation](https://registry.terraform.io/providers/juju/juju/latest/docs). 
```
looks like a typo

orfeas-k (commented):
typo :D

modules/kubeflow-feast/applications.tf
orfeas-k: We discussed this with @NohaIhab and it seems that about kubeflow solution, it was introduced in #28. It was probably a leftover from when we needed to deploy from `1.10/candidate` (see #26). In the future, this could be useful WRT the risk handling story (see #25). However, that's still a WIP, so until then, the `channel` is not needed, let's proceed as is. Maybe the terraform modules store comes first.

orfeas-k (approved):
Good job @NohaIhab :tada:

PR #37, opened 2025-06-04
feat: add deploy Kubeflow+Feast to AKS test and workflow

NohaIhab:
Part of https://github.com/canonical/charmed-kubeflow-uats/issues/180

## Testing
See [this run](https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/15609042213/job/43965294798).
I have tested running the workflow in the CI by overriding the `deploy-to-aks.yaml` workflow in a draft branch from #38 since this is a new workflow I needed a way to override an existing workflow to be able to run it from the actions. #38 will be closed once this PR is merged. 
Note the run is running the UATs from the branch of https://github.com/canonical/charmed-kubeflow-uats/pull/189.

### DIFF
--- a/.github/workflows/deploy-to-aks-feast.yaml
+++ b/.github/workflows/deploy-to-aks-feast.yaml
@@ -88,7 +88,7 @@ jobs:
           cd ~/charmed-kubeflow-uats
           git checkout ${{ env.UATS_BRANCH }}
           # TODO: remove --bundle when feast is promoted to stable 
-          tox -e feast-remote --bundle https://raw.githubusercontent.com/canonical/bundle-kubeflow/refs/heads/main/releases/1.10/stable/bundle.yaml
+          tox -e feast-remote -- --bundle https://raw.githubusercontent.com/canonical/bundle-kubeflow/refs/heads/main/releases/1.10/stable/bundle.yaml
 
       # On failure, capture debugging resources
       - name: Select model
--- a/.github/workflows/deploy-to-aks-feast.yaml
+++ b/.github/workflows/deploy-to-aks-feast.yaml
@@ -1,0 +1,138 @@
+name: Create AKS cluster, deploy kubeflow-feast Terraform solution and run UATs
+on:
+  workflow_dispatch:
+    inputs:
+      k8s_version:
+        description: 'Kubernetes version to be used for the AKS cluster. Make sure that the corresponding K8s version is supported by the cloud.'
+        required: false
+      uats_branch:
+        description: 'Branch to run the UATs from e.g. main or track/1.10. By default, this is defined by the dependencies.yaml file.'
+        required: false
+  schedule:
+    - cron: "17 02 * * 1"
+
+jobs:
+  deploy-solution-to-aks:
+    name: Deploy CKF + Feast solution to AKS
+    runs-on: ubuntu-24.04
+    env:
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-to-aks-feast.yaml
```diff
@@ -0,0 +1,143 @@
...
+      - name: Deploy and assert kubeflow-feast solution
+        run: |
+          tox -c ./modules/kubeflow-feast -vve test_deployment -- -vv -s
+
+      - name: Run UATs
+        run: |
+          git clone https://github.com/canonical/charmed-kubeflow-uats.git ~/charmed-kubeflow-uats
+          cd ~/charmed-kubeflow-uats
+          git checkout ${{ env.UATS_BRANCH }}
+          # TODO: remove --bundle when feast is promoted to stable
```
misohu, line 90: Is this linked in some issue ? if not maybe we should create one so we dont forget.

modules/kubeflow-feast/tests/test_deployment.py
```diff
@@ -0,0 +1,78 @@
...
+        # Verify deployment by checking the public URL
+        url = get_public_url(lightkube_client, "kubeflow")
+        result_status, result_text = await fetch_response(url)
+        assert result_status == 200
+        assert "Log in to Your Account" in result_text
+        assert "Email Address" in result_text
+        assert "Password" in result_text
+
+
+def get_public_url(lightkube_client: lightkube.Client, bundle_name: str):
```
misohu, line 57: Suggestion: Maybe we should use retry here in case there is some temporary problem.

.github/workflows/deploy-to-aks-feast.yaml
```diff
@@ -0,0 +1,143 @@
...
+
+      - name: Run YAML to Github Output Action
+        id: yaml-output
+        uses: christian-ci/action-yaml-github-output@v2
+        with:
+          file_path: ".github/dependencies.yaml"
+
+      - name: Update ENV variables from inputs if available
+        run: |
+          K8S_VERSION=${{ inputs.k8s_version || env.K8S_VERSION }}
```
misohu, line 33: Question (non blocking): I wonder if this `short-circuit evaluation` works. I was always thinking the left side of the evaluation needs to have a logical expression like they mention [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example-matching-an-array-of-strings). But I know we use it also in the different workflow so wonder how it can work haha :)

misohu (commented):
Really nice job :) left few comments

.github/workflows/deploy-to-aks-feast.yaml
NohaIhab: good point, filed https://github.com/canonical/charmed-kubeflow-solutions/issues/39

.github/workflows/deploy-to-aks-feast.yaml
NohaIhab: I tried running it with and without inputs and it is working

NohaIhab: not sure as it contradicts the documentation, but seems to be working correctly

misohu (approved):
Good job

PR #46, opened 2025-06-20
chore: pin jupyter-ui revision and base containing feast notebook image

NohaIhab:
* Adds a `jupyter_ui_base` variable to `kubeflow` solution and uses it in `jupyter_ui` module to set the `base`
* Updates the `kubeflow-feast` solution to set the `jupyter_ui_base` to `24.04` to be able to pull the recent revision containing the feast notebook image
* Pins the value of `jupyter_ui_revision` in `kubeflow-feast` solution to `1209`
* Removes the `jupyter_ui_revision` variable from `kubeflow-feast` solution with a comment

See #47 for details on why we are doing the pinning


NohaIhab:
this is no longer needed, we proceeded with bringing the base migration to stable.
see https://github.com/canonical/notebook-operators/pull/479.

PR #51, opened 2025-06-27
[docs] Installing the starter-pack for CKF documentation migration to RTD 

afgambin:
This PR installs the [starter pack](https://canonical-starter-pack.readthedocs-hosted.com/latest/) as well as required configuration to start the migration of CKF docs to ReadTheDocs.

After agreement, we'll be using this repo as our docs source code


.github/workflows/cla-check.yml
```diff
@@ -0,0 +1,10 @@
+# This workflow checks if the contributor has signed the Canonical Contributor Licence Agreement (CLA)
+name: Canonical Contributor Licence Agreement check
```
orfeas-k, line 2: **question**: Are we sure we need this workflow? I 've seen the check running in repos (e.g. https://github.com/canonical/oidc-gatekeeper-operator/pull/204) where there is no such workflow. Do you have more context?

.github/workflows/automatic-doc-checks.yml
```diff
@@ -0,0 +1,24 @@
+#
+name: Automatic doc checks
+
+on:
+  push:
+    branches: [ main ]
```
orfeas-k, line 6: I think we need this to target `track/` branches as well, specially since right now there is no `main` branch in this repo (see https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.10/.github/workflows/on_push.yaml)

docs/reuse/links.txt
orfeas-k, line 1: I udnerstand files under `docs/reuse` are just placeholders, right?

orfeas-k (commented):
I went through all the files quickly apart from the `/sphix` folder. I understand there is a lot of placeholder plus default configuration here, right? If so, we can proceed once we resolve the comments around CI.

PR #54, opened 2025-07-29
[KF-7695] feat: adding variable and channel pinning

### LINKED ISSUE #52: Have configurable risk for 1.10 solution
### Context

As part of testing https://github.com/canonical/bundle-kubeflow/issues/1277 we need to test the `1.10/edge` bundle. 

But we can't use the TF solution as now it's only using the `stable` risk. https://github.com/canonical/charmed-kubeflow-solutions/pull/25 aimed to make the risk configurable, but was targeted for the 1.9 release.

We'll need the counterpart PR for 1.10, so that SolQA can use these TF solutions for running UATs.

### What needs to get done

1. Introduce a new `risk` variable
2. Modify all KF charms (not mysql) to use this risk variable

Essentially a copy over of https://github.com/canonical/charmed-kubeflow-solutions/pull/25 but for 1.10

### Definition of Done

The 1.10 terraform plan can be configured for which risk should be run

deusebio:
Companion PR for #25 

Closes #52

### DIFF
--- a/modules/kubeflow-feast/applications.tf
+++ b/modules/kubeflow-feast/applications.tf
@@ -3,6 +3,7 @@ module "feast_integrator" {
   source     = "git::https://github.com/canonical/feast-operators//charms/feast-integrator/terraform?ref=track/0.49"
   model_name = module.kubeflow.model
   revision   = var.feast_integrator_revision
+  channel    = "0.49/${var.risk}"
 }
 
 module "feast_ui" {
@@ -10,6 +11,7 @@ module "feast_ui" {
   source     = "git::https://github.com/canonical/feast-operators//charms/feast-ui/terraform?ref=track/0.49"
   model_name = module.kubeflow.model
   revision   = var.feast_ui_revision
+  channel    = "0.49/${var.risk}"
 }
 
 module "feast_offline_store" {
@@ -58,4 +60,5 @@ module "resource_dispatcher" {
   source     = "git::https://github.com/canonical/resource-dispatcher//terraform?ref=track/2.0"
   model_name = module.kubeflow.model
   revision   = var.resource_dispatcher_revision
+  channel    = "2.0/${var.risk}"
 }
--- a/modules/kubeflow-feast/variables.tf
+++ b/modules/kubeflow-feast/variables.tf
@@ -1,3 +1,14 @@
+variable "risk" {
+  type        = string
+  description = "Value for the risk to be used"
+  default     = "stable"
+
+  validation {
+    condition     = contains(["stable", "candidate", "beta", "edge"], var.risk)
+    error_message = "Valid values for var: test_variable are (stable, candidate, beta and edge)."
+  }
+}
+
 variable "cos_configuration" {
   description = "Boolean value that enables COS configuration"
   type        = bool
--- a/modules/kubeflow-feast/variables.tf
+++ b/modules/kubeflow-feast/variables.tf
@@ -5,7 +5,7 @@ variable "risk" {
 
   validation {
     condition     = contains(["stable", "candidate", "beta", "edge"], var.risk)
-    error_message = "Valid values for var: test_variable are (stable, candidate, beta and edge)."
+    error_message = "Valid values for var: risk are (stable, candidate, beta and edge)."
   }
 }
 
--- a/modules/kubeflow-mlflow/main.tf
+++ b/modules/kubeflow-mlflow/main.tf
@@ -1,5 +1,6 @@
 module "kubeflow" {
   source                           = "../kubeflow"
+  risk                             = var.risk
   create_model                     = var.create_model
   cos_configuration                = var.cos_configuration
   dex_connectors                   = var.dex_connectors
@@ -58,8 +59,9 @@ module "kubeflow" {
 }
 
 module "mlflow" {
-  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=track/2.22"
+  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=wip-new-pinning"
   # kubeflow module creates the model
+  risk                        = var.risk
   create_model                = false
   model                       = module.kubeflow.model
   cos_configuration           = var.cos_configuration
--- a/modules/kubeflow-mlflow/main.tf
+++ b/modules/kubeflow-mlflow/main.tf
@@ -59,7 +59,7 @@ module "kubeflow" {
 }
 
 module "mlflow" {
-  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=wip-new-pinning"
+  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=track/2.22"
   # kubeflow module creates the model
   risk                        = var.risk
   create_model                = false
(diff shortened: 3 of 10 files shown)


modules/kubeflow-feast/variables.tf
```diff
@@ -1,3 +1,14 @@
+variable "risk" {
+  type        = string
+  description = "Value for the risk to be used"
+  default     = "stable"
+
+  validation {
+    condition     = contains(["stable", "candidate", "beta", "edge"], var.risk)
+    error_message = "Valid values for var: test_variable are (stable, candidate, beta and edge)."
```
NohaIhab:
```suggestion
    error_message = "Valid values for var: risk are (stable, candidate, beta and edge)."
```

modules/kubeflow-mlflow/variables.tf
```diff
@@ -1,3 +1,14 @@
+variable "risk" {
+  type        = string
+  description = "Value for the risk to be used"
+  default     = "stable"
+
+  validation {
+    condition     = contains(["stable", "candidate", "beta", "edge"], var.risk)
+    error_message = "Valid values for var: test_variable are (stable, candidate, beta and edge)."
```
NohaIhab:
```suggestion
    error_message = "Valid values for var: risk are (stable, candidate, beta and edge)."
```

modules/kubeflow/variables.tf
```diff
@@ -1,3 +1,14 @@
+variable "risk" {
+  type        = string
+  description = "Value for the risk to be used"
+  default     = "stable"
+
+  validation {
+    condition     = contains(["stable", "candidate", "beta", "edge"], var.risk)
+    error_message = "Valid values for var: test_variable are (stable, candidate, beta and edge)."
```
NohaIhab:
```suggestion
    error_message = "Valid values for var: risk are (stable, candidate, beta and edge)."
```

modules/kubeflow-mlflow/main.tf
```diff
@@ -58,8 +59,9 @@ module "kubeflow" {
 }
 
 module "mlflow" {
-  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=track/2.15"
+  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=wip-new-pinning"
```
NohaIhab: suggestion(blocking): pin back to `track/2.15`, assuming this was for testing purposes

NohaIhab (commented):
thanks @deusebio, a much needed change to enable us to move away from the bundle.yaml.

I suggest we add an input to the UATs workflows ([`deploy-to-aks.yaml`](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.10/.github/workflows/deploy-to-aks.yaml) and [`deploy-to-aks-feast.yaml`](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.10/.github/workflows/deploy-to-aks-feast.yaml)) so we can run the tests on different risks.

modules/kubeflow-mlflow/main.tf
NohaIhab: I realize we need to add a `risk` to mlflow solution, hence why you are using the branch. Will review the mlflow PR.

deusebio: yes, I need to update this once the other PR is merged!

modules/kubeflow-mlflow/variables.tf
deusebio: Done!

modules/kubeflow-mlflow/main.tf
deusebio: Done!

modules/kubeflow-feast/applications.tf
```diff
@@ -3,13 +3,15 @@ module "feast_integrator" {
...
 
 module "feast_ui" {
   # tflint-ignore: terraform_module_pinned_source
   source     = "git::https://github.com/canonical/feast-operators//charms/feast-ui/terraform?ref=track/0.49"
   model_name = module.kubeflow.model
   revision   = var.feast_ui_revision
+  channel    = "0.49/${var.risk}"
 }
 
 module "feast_offline_store" {
```
MattiaSarti, line 17: question (non-blocking): Are we intentionally fixing PostgreSQL to `stable` even when our charms parametrize risk? Is it because we see databases as external dependencies with respect to our solutions, that should always work reliably? Because PostgreSQL too [would support all four channel risks for the pinned `14` track](https://charmhub.io/postgresql), if we wanted to.

_This comment applies not only to the **feast_offline_store** module but also to the other PostgreSQL-based modules throughout this file._

MattiaSarti (approved):

modules/kubeflow-feast/applications.tf
deusebio: I would personally use different variables for the risk for internal and external components. And for the scope of the current PR, I would focus and only use risk for the components we control the promotion lifecycle (internal components)

MattiaSarti: Reasonable to me!

afgambin (approved):
lgtm

modules/kubeflow-feast/variables.tf
```diff
@@ -1,3 +1,14 @@
+variable "risk" {
+  type        = string
+  description = "Value for the risk to be used"
+  default     = "stable"
+
+  validation {
+    condition     = contains(["stable", "candidate", "beta", "edge"], var.risk)
```
NohaIhab, line 7: just realized our charms don't have a `<track>/beta` channel for the 1.10 versions, we only have `1.10/beta` channel for the bundle which deploys `latest/beta` charms.

modules/kubeflow-mlflow/main.tf
```diff
@@ -58,8 +59,9 @@ module "kubeflow" {
 }
 
 module "mlflow" {
-  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=track/2.15"
+  source = "git::https://github.com/canonical/charmed-mlflow-solutions//modules/mlflow?ref=track/2.22"
```
NohaIhab, line 62: I agree that this change makes sense, but it feels a bit out of scope for this PR. Would it make more sense to address it in a separate one for better visibility and traceability? I think a change like updating the mlflow component version for the solution should be done in its own PR.

NohaIhab (commented):
a couple comments, we need to figure out our `beta` channels story

modules/kubeflow-feast/variables.tf
NohaIhab: for example see https://charmhub.io/kubeflow-dashboard

modules/kubeflow-mlflow/main.tf
deusebio: Yes, makes sense. Raise https://github.com/canonical/charmed-kubeflow-solutions/pull/61

modules/kubeflow-feast/variables.tf
deusebio: I believe we should have `beta` risk too across the board. But also if we don't have it, I believe when you deploy beta it falls back to the next risk

NohaIhab: you're right, it falls back to `candidate`. Did a quick test:
```
juju deploy kubeflow-dashboard --channel=1.10/beta --trust
Deployed "kubeflow-dashboard" from charm-hub charm "kubeflow-dashboard", revision 815 in channel 1.10/candidate on ubuntu@24.04/stable
```
ok let's not block this then.

NohaIhab (approved):

MattiaSarti (approved):

PR #55, opened 2025-07-31
[KF-7803] Adding Github action for deploying on EKS

afgambin:
This PR addresses this Jira ticket: https://warthogs.atlassian.net/browse/KF-7803

For review: I have been following the same structure we have used for AKS + plus the [Install guide from CKF docs](https://documentation.ubuntu.com/charmed-kubeflow/how-to/install/install-eks/). I am leaving some questions in the PR comments.


.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,130 @@
...
+  MLFLOW_CHANNEL: "latest/stable"
+
+jobs:
+  deploy-solution-to-eks:
+    name: Deploy CKF + MLFlow solution to EKS
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
afgambin: Why do we use `actions/checkout@v2`, instead of @v4 (that seems to be the latest version)? Stability/reproducibility matters?

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,130 @@
...
+      k8s_version:
+        description: 'Kubernetes version to be used for the EKS cluster. Make sure that the corresponding K8s version is supported by the cloud.'
+        required: false
+      uats_branch:
+        description: 'Branch to run the UATs from e.g. main or track/1.10. By default, this is defined by the dependencies.yaml file.'
+        required: false
+  schedule:
+    - cron: "17 02 * * 1"
+
+env:
```
afgambin, line 15: I have defined this `env` field that I do not see in the other workflows. Not sure if we want it this way.

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,130 @@
...
+  CLUSTER_NAME: kubeflow-eks
+  REGION: ${{ secrets.AWS_REGION }}
```
afgambin: We would need to store the AWS credentials + config, e.g., region, as we did for AKS. Can you help me with this? I don't think I have access to those

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,130 @@
...
+
+      - name: Update kubeconfig
+        run: aws eks update-kubeconfig --name $CLUSTER_NAME --region $REGION
+
+      - name: Add EKS cluster to Juju
+        run: |
+          juju version
+          juju add-k8s eks --client
+
+      - name: Deploy and assert kubeflow-mlflow solution
```
afgambin, line 126: I understand from this point below, everything remains the same as for AKS. Do we have to modify something under `./modules/kubeflow-mlflow` folder?

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,130 @@
...
+        run: |
+          tox -c ./modules/kubeflow-mlflow -vve test_deployment -- -vv -s
+
+      - name: Run UATs
```
afgambin, line 134: Some question as above - do we need to update these uats to include EKS? I assume not.

afgambin:
How can I test the workflow works without merging? I mean, I thought to run it manually under Actions tab, but for that, it needs to be already merged

deusebio:
> How can I test the workflow works without merging? I mean, I thought to run it manually under Actions tab, but for that, it needs to be already merged

Just add it to the checks on pull request. Once you make sure it passes, you can disable the workflow on pull_request before merging. 

I was doing this for TIOBE/TICS integration, e.g. [mlflow-operator](https://github.com/canonical/mlflow-operator/pull/320). If you see the history, in the last commit, I just remove the pull request trigger

.github/workflows/deploy-to-eks.yaml
deusebio: I believe we just didn't merge the renovate PR or we don't have automation to handle these updates. Either is fine. I don't think the version for this action is so critical. I would probably use `v2` to be consistent with the other actions, and then raise a PR to fix this consistently in all actions.

.github/workflows/deploy-to-eks.yaml
deusebio: Why not? I don't have strong reason why we should not use this

.github/workflows/deploy-to-eks.yaml
deusebio: The region does not seem a very sensible information, but the other informations (access key and secret key), they surely need to be handled via secrets. I'm not sure I see in the workflow where they are used. I can add these, but can you just give me the information how to call them?

.github/workflows/deploy-to-eks.yaml
deusebio: maybe best for @orfeas-k or @NohaIhab to comment on this

deusebio (commented):
Some first feedbacks and guidance

.github/workflows/deploy-to-eks.yaml
mvlassis: My 2 cents would be to upgrade to `v4` now that we're handling the PR, and potentially in all `checkout` jobs in this repo, since otherwise we are getting deprecation errors. It makes sense to upgrade ASAP since `v2` is deprecated since 2022 according to [this issue](https://github.com/actions/checkout/issues/959).

.github/workflows/deploy-to-eks.yaml
mvlassis: It seems we need 3 variables:
- `AWS_ACCESS_KEY_ID`
- `AWS_ACCESS_SECRET_KEY`
- region

We can use a format similar to our [EKS action on `bundle-kubeflow](https://github.com/canonical/bundle-kubeflow/blob/d33497798cddb14bd70ec01e2406a00b13bcdb68/.github/workflows/deploy-to-eks.yaml#L87-L94), where we inherit these values from GitHub secrets. These are given by a developer (us), in the Repository settings -> Security -> Repository secrets.

To go even further, we can add a variable like AWS_REGION that can be configured in the action itself, although that would probably not offer much.

.github/workflows/deploy-to-eks.yaml
mvlassis: The deployment of the Terraform bundle should be vendor-independent, so we don't need to change anything in this section :)

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,130 @@
...
+        run: |
+          git clone https://github.com/canonical/charmed-kubeflow-uats.git ~/charmed-kubeflow-uats
+          cd ~/charmed-kubeflow-uats
+          git checkout ${{ env.UATS_BRANCH }}
+          tox -e uats-remote -- --filter "not feast"
+
+      # On failure, capture debugging resources
+      - name: Select model
+        run: juju switch aks-controller:kubeflow
```
mvlassis: nitpick: Change the name of the controller (no specific reason for that, just for being clearer), and for all further references to it

Also make sure we actually bootstrap the controller as done [in the AKS action](https://github.com/canonical/charmed-kubeflow-solutions/blob/716ffa1cdf0c552baffe5b037295df0fde5997bb/.github/workflows/deploy-to-aks.yaml#L75-L79)
```suggestion
        run: juju switch eks-controller:kubeflow
```

.github/workflows/deploy-to-eks.yaml
afgambin: > I would probably use `v2` to be consistent with the other actions, and then raise a PR to fix this consistently in all actions.

Sounds the best approach to me here. I'll keep v2 then in this PR. I have created an upgrade issue for future work: https://github.com/canonical/charmed-kubeflow-solutions/issues/59

.github/workflows/deploy-to-eks.yaml
afgambin: Yes, what Manos says. I thought of replicating the EKS action, i.e., using GitHub secrets for this. I have updated the .yaml adding these as env vars:
AWS_REGION: ${{ secrets.AWS_REGION }}
AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}

.github/workflows/deploy-to-eks.yaml
afgambin: good catch! Fixed

.github/workflows/deploy-to-eks.yaml
deusebio: Perfect. Adding these then!

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,134 @@
...
+    - cron: "17 02 * * 1"
+
+env:
+  CLUSTER_NAME: kubeflow-eks
+  AWS_REGION: ${{ secrets.AWS_REGION }}
+  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  JUJU_CONTROLLER: kubeflow-controller
+  JUJU_MODEL: kubeflow
+  CKF_CHANNEL: "1.10/stable"
```
deusebio: **todo** this would really seem an input for the workflow, such that we can use this to also test other releases. A bit along side what we do in the bundle testing [here](https://github.com/canonical/bundle-kubeflow/blob/main/.github/workflows/deploy-to-eks.yaml#L4). If we want to use the TF in a centered way, I feel we need to allow for this customization/parametrizazione of the workflows. 

I would probably do an improvement with respect to the bundle tests, where we could use two parameters here:
* track, e.g. 1.10 (which select the branch here to be used, e.g. `track/1.10`)
* risk, e.g. `candidate`, which could use the risk variable as introduced in #54 

We can also address this in a separate PR where we do this change for both AKS and EKS, also given the risk PR still needs to land. If so, could you create an issue and post it here below, and address on a separate PR? But, if you prefer to address this on in this PR, that's very fine too

.github/workflows/deploy-to-eks.yaml
deusebio: **todo, out-of-scope** Yes, this sounds good! However, I would prioritize this work straight after this PR is merged @afgambin

deusebio (commented):
I have added the credentials and re-triggered the CI. It seems that it is progressing through the "Configure AWS credentials" and it is currently running the "Create EKS cluster" (finger crossed). Please keep an eye and check if the tests work fine, otherwise update the PR accordingly.

Overall, I believe the action here is **almost** good (we are missing the cleanup of the EKS cluster steps, [here](https://github.com/canonical/bundle-kubeflow/blob/main/.github/workflows/deploy-to-eks.yaml#L185) - also with the deletion of the volumes). However I have pointed a couple of improvements (v2 migration to v4, and adding inputs to allow testing of multiple releases), that I believe we can also address as separate ticket to keep well-perimetered PRs, and avoid scope creep. 

However I feel this are important points that we could address straight away, after this PR is merged, still within the scope of the current epic.

deusebio (changes requested):
unfortunately, the action seemed to failed. I checked the logs and there seems to possibly be an [permission error](https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/16719530591/job/47352273016?pr=55#step:9:22) (sometimes I saw this because of the confinemtn of the snap, which therefore does not find/see the aws executable). My suggesiton would be to also use the other action [here](https://github.com/canonical/bundle-kubeflow/blob/main/.github/workflows/deploy-to-eks.yaml#L102) to take inspiration.

More worringly, the current CI seems to miss the cleanup stanges [here](https://github.com/canonical/bundle-kubeflow/blob/main/.github/workflows/deploy-to-eks.yaml#L185). Indeed, I can still see the dangling resources attached, which I'm now clearing manually, but make sure that you cover these steps

.github/workflows/deploy-to-eks.yaml
NohaIhab: good point, I'm +1 to address this in a separate PR where we introduce the inputs for both AKS and EKS workflows

NohaIhab: on the `track` input, wouldn't this be the same as selecting the branch from which to run the workflow? since the solution lives in the same repo as the workflow. In that case we don't need a `track` input.

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,134 @@
...
+  MLFLOW_CHANNEL: "latest/stable"
```
NohaIhab: question: where does this reflect in the solution? I don't see a use of this env atm. The mlflow solution gets deployed with the branch in the `source` [here](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.10/modules/kubeflow-mlflow/main.tf#L61).

.github/workflows/deploy-to-eks.yaml
NohaIhab: question: same as the `MLFLOW_CHANNEL` env, where are these envs being used?
my suggestion for the `JUJU_CONTROLLER` and `JUJU_MODEL` is to remove them, there's no need to parameterize those. For the `CKF_CHANNEL` let's decide it in the other discussion.

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+      - name: Install CLI tools & dependencies
+        run: |
+          pip install tox
+          sudo snap install juju --channel=${{ env.JUJU_VERSION }}/stable
+          sudo snap install charmcraft --classic
+          sudo snap install terraform  --channel latest/stable --classic
+          juju version
+          terraform --version
+          curl --silent --location "https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_$(uname)_amd64.tar.gz" | sudo tar xz -C /usr/local/bin
```
NohaIhab: suggestion: looks like the `eksctl` tool is now moved to its own GH org, see [latest docs](https://eksctl.io/installation/). The `weaveworks` one has a redirect to the new one `eksctl-io`, so let's just use the latest one.

NohaIhab (commented):
thanks @afgambin, some comments from my side

.github/workflows/deploy-to-eks.yaml
afgambin: Task created: https://github.com/canonical/charmed-kubeflow-solutions/issues/60

.github/workflows/deploy-to-eks.yaml
afgambin: Good point - refactored it so we use the `test`file as source of truth for this

.github/workflows/deploy-to-eks.yaml
afgambin: I have removed then

.github/workflows/deploy-to-eks.yaml
afgambin: Updated!

afgambin:
@NohaIhab @mvlassis, maybe also @deusebio if you have the chance. I have been addressing your comments and tweaking the solution. A summary of the current status:
1. I'm afraid that there are some hanging resources from the first version I run, that I cannot eliminate and are making the action to fail. I added pre-deletion sections for the EKS cluster and CloudFormation, but I am getting an error for the `kubeflow`namespace. I have added also another run to clean up the namespace, but the error still persists. I am somehow blocked here since I cannot manually run this or access the AWS console to see what's there and remove resources - could you help me with that?
2. I have pinned certain versions, for Python and Juju, because I was getting errors from `tox` dependencies.
3. Same story with `pytest`, I had to manually installing it, even though it is defined as part of  `testenv:test_deployment` in `tox.ini`. Not sure why.

I understand points 2 and 3 are not ideal (pinning versions manually) but not sure where we set those. Maybe in the UATs repo?

deusebio:
@afgambin I believe I have deleted all the resources. 

If this still doesn't work and we keep having dangling resources, we should probably place your credentials for the time being into the secrets, such that you can debug and not be blocked by me. Once the PR lands, we can swap the credentials again. 

I see that the errors right now are terraform related, so probably we need to investigate the business logic

deusebio:
Just checked, and yes, I believe it is business logic related. 

The terraform is creating the juju_model (see [here](https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/16779498340/job/47514016628?pr=55#step:17:2110)), since the default value for the `create_model` variable is `true` (see [here](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.10/modules/kubeflow-mlflow/variables.tf#L4)). But you also create the model with the CLI in the CI, see [here](https://github.com/canonical/charmed-kubeflow-solutions/pull/55/files#diff-83526f9be343160cff35d00d035545c0db8d19a416e5190f3eb26e12500dea64R124), hence the process complaining that the model already exists. 

You probably just need to remove that line from the CI (or set the `create-model` variable to `false`, but I believe this requires more work)

.github/workflows/deploy-to-eks.yaml
NohaIhab: we can discuss the input in the relevant issue, thanks for creating it @afgambin 
resolving

deusebio:
Since I have seen that there has been new runs, I have noticed that the CI correctly deletes most of the resources, but not the volumes. Should we look into this?

See the CI in the bundle [here](https://github.com/canonical/bundle-kubeflow/blob/main/.github/workflows/deploy-to-eks.yaml#L200)

afgambin:
PR status update:
- I have added the workflow for deleting AWS modules - replicating what we do in the bundle-kubeflow repo.
- AWS region is no longer needed as a GH secret. @deusebio you can eliminate it for this repo. Now, it is extracted from cluster.yaml - replicating what we are doing in the bundle-kubeflow repo.
- It seems when testing the solution ("Deploy ..." step), there is an issue with the `minio-mlflow` resource. Investigating it. That step is also broken in the other action (deploying CKF+Feast to AKS). In this case, this is a summary of what I have seen:

During CI runs, the Terraform deployment of mlflow-minio consistently stalled. Potential causes:
- The mlflow-minio charm requires an OCI image resource to start.
- The Terraform module for mlflow-minio is not defined in our codebase: it is brought in as an indirect dependency from an upstream module.
- Because our code does not explicitly attach the required oci-image resource, Juju reported the application as blocked.

afgambin:
I have been testing a couple of things but the error still remains. This seems to be the issue:

Terraform is trying to deploy a `minio` application inside the `mlflow` module. The minio charm referenced in that module doesn’t point to a valid OCI image.That definition comes from upstream, so we don’t control its OCI image.
 As a result, Juju keeps the application in “creating…” state indefinitely.

Not sure how to solve this, at least not for a long-term solution. @NohaIhab @mvlassis

.github/workflows/deploy-to-aks.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
       - name: Install CLI tools
         run: |
           pip install tox
-          sudo snap install juju --channel=${{ env.JUJU_VERSION }}/stable
+          sudo snap install juju --channel=${{ env.JUJU_VERSION }}/latest/stable
```
mvlassis: issue: I think we will encounter an issue here, since the channel `3.6/latest/stable` doesn't exist. I think you probably meant to keep `3.6/stable`. This command would return:
```
sudo snap install juju --channel=3.6/latest/stable
Warning: Specifying a channel "3.6/latest/stable" is relying on undefined
         behaviour. Interpreting it as "3.6/latest/stable" for now, but this
         will be an error later.

error: invalid risk in channel name: 3.6/latest/stable
```
```suggestion
          sudo snap install juju --channel=${{ env.JUJU_VERSION }}/stable
```

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,205 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install CLI tools & dependencies
+        run: |
+          pip install tox
+          sudo snap install juju --channel=${{ env.JUJU_VERSION }}/stable
+          sudo snap install charmcraft --classic
+          sudo snap refresh charmcraft --channel=3.x/stable
```
mvlassis: nitpick: One-liner suggestion:
```suggestion
          sudo snap install charmcraft --channel=3.x/stable --classic
```

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,205 @@
...
+          sudo snap install terraform --channel=latest/stable --classic
```
mvlassis: nitpick: One-liner suggestion:
```suggestion
          sudo snap refresh charmcraft --channel=3.x/stable
          sudo snap install terraform --channel=latest/stable --classic
```

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,204 @@
...
+          echo "UATS_BRANCH=${UATS_BRANCH}" >> $GITHUB_ENV
+
+      - name: Extract AWS region from cluster.yaml
+        id: extract_region
+        run: |
+          REGION=$(yq e '.metadata.region' .github/cluster.yaml)
+          echo "AWS_REGION=$REGION" >> $GITHUB_ENV
+          echo "region=$REGION" >> $GITHUB_OUTPUT
+
+      # Double-check if this is still needed - same with pinning Charmcraft version
```
afgambin: I was getting errors related to Python and Charmcraft versions at the beginning. Not sure if that's still the case, i.e., maybe the error was coming from something else. I'll try to rerun without pinning these versions.

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,204 @@
...
+          kubectl version --client
+
+      - name: Install eksctl
+        run: |
+          PLATFORM=$(uname -s)_amd64
+          curl -sL "https://github.com/eksctl-io/eksctl/releases/latest/download/eksctl_${PLATFORM}.tar.gz" | tar xz -C /tmp
+          sudo mv /tmp/eksctl /usr/local/bin
+          eksctl version
+
+      # Once working, do we want to keep these two pre-deletion steps?
```
afgambin, line 83: Once working, do we want to keep these two pre-deletion steps? I think they’re a good way to avoid a mess-up, like what happened the first time I ran the action. But if we’re sure it works, do we still need the steps?

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,204 @@
...
+          K8S_VERSION=${{ inputs.k8s_version || env.K8S_VERSION }}
+          echo "K8S_VERSION=${K8S_VERSION}" >> $GITHUB_ENV
+          UATS_BRANCH=${{ inputs.uats_branch || env.UATS_BRANCH }}
+          echo "UATS_BRANCH=${UATS_BRANCH}" >> $GITHUB_ENV
+
+      - name: Extract AWS region from cluster.yaml
+        id: extract_region
+        run: |
+          REGION=$(yq e '.metadata.region' .github/cluster.yaml)
+          echo "AWS_REGION=$REGION" >> $GITHUB_ENV
```
afgambin, line 46: The AWS region is no longer obtained from the GH secret. That secret needs to be removed (I cannot because of permissions).

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,204 @@
...
+        description: 'Kubernetes version to use for the EKS cluster (e.g. 1.27)'
+        required: false
+      uats_branch:
+        description: 'Branch to run the UATs from, e.g., main or track/1.10'
+        required: false
+  schedule:
+    - cron: "17 02 * * 1"
+
+env:
+  CLUSTER_NAME: kubeflow-eks2
```
afgambin: When running this with `kubeflow-eks` as the cluster name, an error occurs. Saying "the cluster already exists", even though I believe we removed all of it properly. So I renamed it like this.

.github/workflows/deploy-to-eks.yaml
NohaIhab: I'd rather we keep it yes, as a fail-safe for example if the deletion in the previous run did not work

.github/workflows/deploy-to-eks.yaml
NohaIhab: I believe there was a leftover cluster that is now removed, you can check with the team's credentials `eksctl get cluster --region eu-central-1`. If it's not there we can remove the `2`.

.github/workflows/deploy-to-eks.yaml
afgambin: sgtm

.github/workflows/deploy-to-eks.yaml
afgambin: I renamed it eventually to `kubeflow-eks-test`, so the purpose of the cluster is clearer.

.github/cluster.yaml
```diff
@@ -0,0 +1,38 @@
...
+kubernetesNetworkConfig:
+  ipFamily: IPv4
+managedNodeGroups:
+- amiFamily: Ubuntu2204
+  iam:
+    withAddonPolicies:
+      ebs: true
+  instanceType: t2.2xlarge
+  labels:
+    alpha.eksctl.io/cluster-name: scrumptious-wardrobe-1684842095
```
NohaIhab: I think this label would be auto-generated based on the metadata.name. While it works fine, it can cause confusion with handling the resources. Most probably it is currently being ignored by `eksctl`. I suggest we remove it from the cluster.yaml.

.github/cluster.yaml
```diff
@@ -0,0 +1,38 @@
...
+    alpha.eksctl.io/nodegroup-name: ng-d06bd84e
```
NohaIhab: This is also auto-generated as a randomly generated nodegroup ID, I suggest we remove it.

.github/cluster.yaml
NohaIhab: you can test by running `eksctl create cluster --name kubeflow-eks-test --dry-run`

NohaIhab (commented):
thanks @afgambin, great job! just a small comment on the cluster labels, probably left-over from previous workflow

NohaIhab (approved):
thank you @afgambin, truly valuable contribution!

afgambin:
@deusebio this is eventually passing all checks after Noha unblocked it, solving the issue with the mlflow-minio resource. Please have a look whenever you can for final approval, ty!

.github/cluster.yaml
```diff
@@ -0,0 +1,35 @@
...
+  ssh:
+    allow: true
+  tags:
+    alpha.eksctl.io/nodegroup-name: ng-d06bd84e
+    alpha.eksctl.io/nodegroup-type: managed
+  volumeSize: 100
+metadata:
+  name: kubeflow-test
+  region: eu-central-1
+  version: "1.26"
```
deusebio: **edit** is this the K8s version? this should be 1.32

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,198 @@
...
+
+      - name: Set envvars from dependencies.yaml
+        run: |
+          yq eval 'to_entries | .[] | "\(.key)=\(.value)"' ".github/dependencies.yaml" | while IFS= read -r line; do
+            echo "$line" >> "$GITHUB_ENV"
+          done
+
+      - name: Update ENV variables from inputs if available
+        run: |
+          K8S_VERSION=${{ inputs.k8s_version || env.K8S_VERSION }}
```
deusebio, line 37: **question** where is this used? the `env` is sourced from dependencies.yaml

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,198 @@
...
+        run: |
+          mkdir -p ~/.aws
+          aws configure set aws_access_key_id "${{ secrets.AWS_ACCESS_KEY_ID }}"
+          aws configure set aws_secret_access_key "${{ secrets.AWS_SECRET_ACCESS_KEY }}"
+          aws configure set default.region "${{ env.AWS_REGION }}"
+          echo "AWS_SDK_LOAD_CONFIG=1" >> "$GITHUB_ENV"
+
+      - name: Install kubectl
+        run: |
+          sudo snap install kubectl --classic --channel=${{ env.K8S_VERSION }}/stable
```
deusebio, line 72: We need to make sure that the K8s version here aligns with the one of the cluster being deployed. If any, the most important versions to be specified is the one of the cluster. It could still be that the API is consistent, so even when the two mismatch, the `kubectl` commands still work, but we should keep them aligned.

deusebio (changes requested):
There is something fuzzy/inconsistent with the K8s version I feel. But overall, by default, we should test on 1.32, which is the LTS

.github/workflows/deploy-to-eks.yaml
afgambin: I agree. But shouldn’t the version specified in `dependencies.yaml` be the source of truth here?
That’s the one we pass to kubectl:

`sudo snap install kubectl --classic --channel=${{ env.K8S_VERSION }}/stable`

And it’s also the one that prevails when creating the EKS cluster, even if there’s a mismatch. If a mismatch occurs during creation, the version in dependencies.yaml wins:

`yq e ".metadata.version |= \"${{ env.K8S_VERSION }}\"" -i .github/cluster.yaml`

.github/cluster.yaml
afgambin: updated

.github/workflows/deploy-to-eks.yaml
afgambin: We use K8S_VERSION for kubectl and creating the EKS cluster. This step was added for potential overrides when manually triggering the action (same pattern we had in bundle-kubeflow).

.github/workflows/deploy-to-eks.yaml
```diff
@@ -0,0 +1,198 @@
+name: Create EKS cluster, deploy kubeflow-mlflow Terraform solution and run UATs
+
+on:
+  pull_request: {}
```
afgambin: btw, do we want to remove this or keep triggering the action for every PR?

deusebio: Yes. Let's remove, we should just have it scheduled and manually dispatched

.github/workflows/deploy-to-eks.yaml
deusebio: Oh I see. Sounds good!

deusebio (approved):
Looks good! Thanks @afgambin !!

PR #76, opened 2025-10-08
feat: add registration flow config for kubeflow dashboard

toaksoy:


modules/kubeflow/variables.tf
```diff
@@ -331,6 +331,12 @@ variable "kubeflow_dashboard_revision" {
   default     = null
 }
 
+variable "kubeflow_dashboard_registration_flow" {
+  description = "Whether to enable the registration flow on sign-in for kubeflow-dashboard"
+  type        = string
+  default     = "true"
```
deusebio, line 337: *note* Looks good ! the value is consistent with the default of the charm, so this should not change current behaviour

deusebio (approved):
Looks good!

modules/kubeflow/variables.tf
toaksoy: thanks for the quick review @deusebio

MattiaSarti: True, I can confirm [so](https://github.com/canonical/kubeflow-dashboard-operator/blob/main/config.yaml#L39).

MattiaSarti (approved):

deusebio:
It is fine to merge in RED. The AKS tests won't work for forks

deusebio:
@AxoyTO could you sign the commits? we can't merge as the commits have not been signed

MattiaSarti (approved):

PR #77, opened 2025-10-08
[KF-7944] Adding K8s bundle tests

### LINKED ISSUE #67: Create action for deploying Terraform solution and asserting it deploys successfully
### Context

To ensure that changes to this repository don't create any issues, we should create a GitHub action that:
- Uses the self-hosted runners
- Deploys the Terraform solution (the channel should correspond to the branch, e.g. `track/1.10` should deploy `1.10/stable`
- Asserts that the solution is deployed successfully

### What needs to get done

- Look into whether we should just deploy the `kubeflow` solution, or potentially `kubeflow-mlflow` and `kubeflow-feast` as well
- Use Canonical K8s 

### Definition of Done

.

deusebio:
Addresses [KF-7944](https://warthogs.atlassian.net/browse/KF-7944)

Fixes https://github.com/canonical/charmed-kubeflow-solutions/issues/67

[KF-7944]: https://warthogs.atlassian.net/browse/KF-7944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/concierge.yaml
```diff
@@ -0,0 +1,25 @@
...
+      load-balancer:
+        enabled: true
+        l2-mode: true
+        cidrs: 10.64.140.43/32
+    bootstrap-constraints:
+      root-disk: 4G
+
+  lxd:
+    enable: true
+    bootstrap: false
```
NohaIhab, line 22: suggestion(non-blocking): set the channel for lxd to `latest/stable`

.github/concierge.yaml
```diff
@@ -0,0 +1,25 @@
+juju:
+  channel: 3.6/stable
```
NohaIhab, line 2: suggestion(non-blocking): increase logging level:

```suggestion
  channel: 3.6/stable
  model-defaults:
    logging-config: <root>=INFO; unit=DEBUG
```

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -0,0 +1,149 @@
...
+          sudo apt-get remove -y docker.io containerd
+          sudo rm -rf /run/containerd
+          sudo snap install concierge --classic
+          cd .github
+          
+          # Set the K8s version
+          yq -i e '.providers.k8s.channel = "${{ env.K8S_VERSION }}-classic/stable"' concierge.yaml
+          
+          sudo concierge prepare --trace
+          cd ..
```
dariofaccin, line 97: Is there a particular reason to have the `concierge.yaml` fil inside the `.github` folder?

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -0,0 +1,149 @@
...
+          sudo snap install terraform  --channel latest/stable --classic
+          terraform --version
+
+      - name: Deploy solution
+        run: |
+          if [[ '${{ github.event_name }}' != 'pull_request' ]]; then
+            git checkout '${{ env.CKF_BRANCH }}'
+          fi
+
+          tox -c ./modules/${{ matrix.bundle.module }} -vve test_deployment -- -vv -s --risk ${{ env.RISK }}
```
dariofaccin: Why `-vv` for `tox`?

dariofaccin (commented):
LGTM, I have a couple of questions before approval.

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -0,0 +1,149 @@
...
+          sudo mkdir -p ${MIRROR_CONFIG}
+          sudo chown $USER ${MIRROR_CONFIG}
+          cat << EOF > ${MIRROR_CONFIG}/hosts.toml
+          [host."$DOCKERHUB_MIRROR"]
+          capabilities = ["pull", "resolve"]
+          EOF
+          fi
+          
+          sudo apt-get remove -y docker.io containerd
+          sudo rm -rf /run/containerd
```
NohaIhab, line 89: question: why is this needed?

modules/kubeflow-feast/tests/conftest.py
```diff
@@ -33,8 +33,33 @@ def pytest_addoption(parser):
         type=str,
         help="Risk to be used when deploying the terraform module",
     )
+    parser.addoption(
+        "--db-size",
+        default="1G",
+        type=str,
+        help="Size to be used for the databases.",
+    )
```
NohaIhab, line 41: question: why add an arg for db size?

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -15,7 +15,6 @@ on:
         description: 'The risk level to be tested'
         required: false
         default: 'stable'
-
   pull_request:
```
NohaIhab, line 18: should we remove the on pull request from this one? the Ck8s should be replacing it IMO

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -0,0 +1,149 @@
...
+        description: 'The risk level to be tested'
+        required: false
+        default: 'stable'
+  pull_request:
+  schedule:
+    - cron: "17 02 * * 1"
+
+jobs:
+  deploy-solution:
+    name: Deploy CKF + MLFlow solution to K8s
```
NohaIhab: issue: the name of the job for CKF + Feast is now saying CKF + MLFlow

NohaIhab (commented):
thanks @deusebio, really nice seeing it pass in the CI! Left some comments, open to discussion.

.github/workflows/deploy-to-k8s.yaml
deusebio: This file is just used to setup the environment in CI. This would represent similar parameters as the ones defined [here](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.10/.github/workflows/deploy-to-microk8s.yaml#L76) for microk8s. As such I felt this information was part of CI stuff, hence the position into `.github` folder. 

But I don't have very strong opinion if we have better suggestions or thoughts.

.github/workflows/deploy-to-k8s.yaml
deusebio: I believe the `-vv` is for pytest, to provide more verbosity, see [here](https://docs.pytest.org/en/stable/how-to/output.html#verbosity). This is the way we have always run tests, and not very specific to this CI

.github/workflows/deploy-to-k8s.yaml
deusebio: I believe I brought this up during one of our meeting: this is due to the ratelimiting that self-hosted runners might be subjected to, see [here](https://canonical-self-hosted-github-runner-docs.readthedocs-hosted.com/en/latest/usage/faq/how-to-avoid-dockerhub-rate-limits/)

modules/kubeflow-feast/tests/conftest.py
deusebio: On Canonical K8s we need to limit the size of the storage, or the pvc are not allocated. I believe it may be that Canonical K8s either consumes more disk space or has some restriction on how to use the disk space of the hosting machine. Either way, the deployment would hang if the storage were 10GB because juju could not attach any more storage after the first few volumes

.github/workflows/deploy-to-microk8s.yaml
deusebio: I believe it is fair to have both honestly since we still provide our tutorial using Microk8s, and it would be good to see if there are any regressions. Beside having a bit more coverage wouldn't hurt us, but yes, in the mid- or long- term, I believe we only want to keep Canonical K8s

.github/workflows/deploy-to-k8s.yaml
deusebio: renamed to simply "Deploy to ..." for all 4 workflows on EKS, AKS, MicroK8s and Canonical K8s

.github/workflows/deploy-to-k8s.yaml
dariofaccin: Yes, we had to use the same approach in my previous team to avoid the pull limit. Maybe a comment referencing the docs would be nice here for context.

.github/workflows/deploy-to-k8s.yaml
dariofaccin: I was referring to this part: `tox -c ./modules/${{ matrix.bundle.module }} -vve test_deployment`.  From the logs of [this run](https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/19482645233/job/55788616132?pr=77#step:10:15), lines 15 to 89 are debug logs of `tox` itself.

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -0,0 +1,149 @@
...
+          sudo snap install terraform  --channel latest/stable --classic
+          terraform --version
+
+      - name: Deploy solution
+        run: |
+          if [[ '${{ github.event_name }}' != 'pull_request' ]]; then
+            git checkout '${{ env.CKF_BRANCH }}'
+          fi
+
+          tox -c ./modules/${{ matrix.bundle.module }} -vve test_deployment -- -vv -s --risk ${{ env.RISK }}
```
dariofaccin:
```suggestion
          tox -c ./modules/${{ matrix.bundle.module }} -e test_deployment -- -vv -s --risk ${{ env.RISK }}
```

.github/workflows/deploy-to-k8s.yaml
deusebio: sure! I'll add that!

.github/workflows/deploy-to-k8s.yaml
deusebio: I have removed the `-vv`, as I don't feel very strongly on this, but I kinda copy and pasted from other tests, where we generally use that syntax. I have now removed it on [3fc2c53](https://github.com/canonical/charmed-kubeflow-solutions/pull/77/commits/3fc2c53a3750f7fc395a5f7552950fa28577c03b). 

Please @NohaIhab chime in if people really cared about this.

.github/workflows/deploy-to-k8s.yaml
deusebio: Done on [3fc2c53](https://github.com/canonical/charmed-kubeflow-solutions/pull/77/commits/3fc2c53a3750f7fc395a5f7552950fa28577c03b).

.github/workflows/deploy-to-k8s.yaml
dariofaccin: I agree, my question arose since in all other repos we have that file at root level and I would keep the same structure (that also allows to remove the two `cd` commands).

dariofaccin: Agreed to maintain the file under `.github` folder.

dariofaccin (approved):
LGTM, assuming CI issues are temporary.

.github/workflows/deploy-to-microk8s.yaml
NohaIhab: fair enough, let's keep it for now then and re-evaluate later

NohaIhab (approved):
thanks @deusebio

deusebio:
The tests took a bit too many re-run to pass for my taste. It seems that the flakiness has somewhat increases since when I raised the PR. I believe we should look into this, but I'd merge this PR to also unblock Dario on his work

PR #78, opened 2025-10-09
Remove configs that are nulls

### LINKED ISSUE #72: Remove annotations config option from track/1.9 - breaks deployment
The changes in commit 062665b fix some issues but have also introduced the `annotations` option to istio-ingressgateway. This option doesn't exist in channel 1.22/stable which is the default channel selected in the terraform module: https://github.com/canonical/istio-operators/blob/track/1.22/charms/istio-gateway/config.yaml

Could you please either:
a) Remove this option, or make it conditional on the channel
b) Set the default of the annotations to `null` and not `""`.

The second point b) should be followed in any case. Config options to juju charms should never be specified with their defaults, they should use `null`. Otherwise the juju application will say `source: user` rather than `source: default`. In addition, it means that the default configs will need to be updated in two locations: the charm and the terraform module, rather than just one place.

pengwyn:
Fixes #72

This is one idea to try and handle nulls in the config. However, alternatives might be better:
- removing the config altogether (but this would not allow it to be added in later)
- using map merges instead of a filter
- a more standardised method for `juju_application` configs that I'm probably not aware of.

### DIFF
--- a/modules/kubeflow/applications.tf
+++ b/modules/kubeflow/applications.tf
@@ -33,13 +33,19 @@ module "envoy" {
   channel    = "2.2/${var.risk}"
 }
 
+locals {
+  istio_ingressgateway_config = {
+    kind        = "ingress",
+    annotations = var.istio_ingressgateway_annotations,
+  }
+}
 module "istio_ingressgateway" {
   source     = "git::https://github.com/canonical/istio-operators//charms/istio-gateway/terraform?ref=track/1.22"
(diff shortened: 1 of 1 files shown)


deusebio:
This is not needed anymore, since the TF provider (at least at version 0.23.0) has been shown to support and handle null config (see discussion [here](https://github.com/canonical/charmed-kubeflow-solutions/pull/80#issuecomment-3388408734))

Issue #74, opened 2025-10-08
mysql-k8s charms in error "Unable to get member status"


motjuste:
### Bug Description

After recent updates for the mysql terraform changes, the current 1.10 plan does not deploy with all the mysql-k8s charms in error state saying: "Unable to get member state".

### To Reproduce

SQA deploying the Terraform plan from track/1.10 on AKS

### Environment

Solutions QA

### Relevant Log Output

```shell
$ juju status -m kubeflow
Model     Controller      Cloud/Region      Version  SLA          Timestamp
kubeflow  aks-controller  aks-cloud/eastus  3.6.9    unsupported  12:59:50Z

App                      Version                Status       Scale  Charm                    Channel           Rev  Address       Exposed  Message
admission-webhook                               active           1  admission-webhook        1.10/stable       467  10.0.78.113   no
argo-controller                                 active           1  argo-controller          3.5/stable        818  10.0.135.159  no
dex-auth                                        active           1  dex-auth                 2.41/stable       699  10.0.73.16    no
envoy                                           active           1  envoy                    2.4/stable        406  10.0.176.149  no
istio-ingressgateway                            active           1  istio-gateway            1.24/stable      1474  10.0.88.76    no
istio-pilot                                     active           1  istio-pilot              1.24/stable      1417  10.0.188.206  no
jupyter-controller                              active           1  jupyter-controller       1.10/stable      1260  10.0.123.176  no
jupyter-ui                                      active           1  jupyter-ui               1.10/stable      1222  10.0.199.218  no
katib-controller                                active           1  katib-controller         0.18/stable      1163  10.0.178.212  no
katib-db                                        maintenance      3  mysql-k8s                8.0/stable        255  10.0.245.88   no       Unable to get member state
katib-db-manager                                waiting          1  katib-db-manager         0.18/stable      1123  10.0.165.83   no       Incorrect/incomplete data found in relation relational-db. See logs
katib-ui                                        active           1  katib-ui                 0.18/stable      1122  10.0.139.137  no
kfp-api                                         waiting          1  kfp-api                  2.5/stable       2240  10.0.196.62   no       Incorrect/incomplete data found in relation relational-db. See logs
kfp-db                                          maintenance      3  mysql-k8s                8.0/stable        255  10.0.56.71    no       Unable to get member state
kfp-metadata-writer                             active           1  kfp-metadata-writer      2.5/stable       1310  10.0.24.213   no
[...]

PR #80, opened 2025-10-09
Revert "rollback: removing annotation to istio-gateway (#79)"

deusebio:
This reverts commit d2f03f5081e5a2939ebb5a083425e05fcff88bce.

This is to be tested by ManSol

### DIFF
--- a/modules/kubeflow-mlflow/main.tf
+++ b/modules/kubeflow-mlflow/main.tf
@@ -23,6 +23,7 @@ module "kubeflow" {
   dex_auth_revision                = var.dex_auth_revision
   envoy_revision                   = var.envoy_revision
   istio_ingressgateway_revision    = var.istio_ingressgateway_revision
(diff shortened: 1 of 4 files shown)


pengwyn:
Tested and it works! For completeness I changed our TF to target 
```
      version = ">= 0.23.0, < 0.24"
```
and this was enough to handle the `null` config options. We have updated our repos for new deployments, so they should be able to handle this change if it were to land now. Having said that, older deployments would fail if some changes were to be added, but I believe we can cross that bridge when we come to it.

Thanks for taking care of this.

deusebio:
Yes, that's fine. I suppose older deployments are nevertheless broken because of #74 , so they would need to upgrade either way. While upgrading they can also update the TF provider. 

As soon as someone from the team will approve, we will merge the PR

MattiaSarti (approved):

Issue #60, opened 2025-08-05
Add workflow release inputs for AKS and EKS


afgambin:
### Context

Release version could be an input for the EKS and AKS workflows, such that we can use it to test other releases. Similar to what we do in the bundle testing [here](https://github.com/canonical/bundle-kubeflow/blob/main/.github/workflows/deploy-to-eks.yaml#L4).
This would enable a better workflow customization. 

We could use two parameters here:

   - track, e.g. 1.10 (which select the branch here to be used, e.g. track/1.10).
   - risk, e.g. candidate, which could use the risk variable as introduced in [here](https://github.com/canonical/charmed-kubeflow-solutions/pull/54).

### What needs to get done

Add release input to EKS and AKS workflows.

### Definition of Done

The input has been added to both GH actions.

PR #81, opened 2025-10-10
[KF-7841] Add workflow release inputs

deusebio:
Addressing #60


deusebio:
Ideally, we could prioritize #82 and then add this features of controlling release inputs to those workflows as well

modules/kubeflow-mlflow/tests/test_deployment.py
```diff
@@ -1,4 +1,5 @@
 import subprocess
+from email.policy import default
```
mvlassis: question: I'm not sure where this import is used
```suggestion
from email.policy import default
```

.github/workflows/deploy-to-aks.yaml
```diff
@@ -81,7 +96,11 @@ jobs:
 
       - name: Deploy and assert kubeflow-mlflow solution
         run: |
-          tox -c ./modules/kubeflow-mlflow -vve test_deployment -- -vv -s
+          if [[ '${{ github.event_name }}' != 'pull_request' ]]; then
+            git checkout 'track/${{ env.CKF_VERSION }}'
+          fi
```
mvlassis, line 116: question: I'm not sure why we only `checkout` on pull requests. If I understand correctly, this workflow only runs on workflow dispatch, so this never occurs.
```suggestion
          if [[ '${{ github.event_name }}' != 'pull_request' ]]; then
            git checkout 'track/${{ env.CKF_VERSION }}'
          fi
```

mvlassis (commented):
Left small comments-questions

.github/workflows/deploy-to-aks.yaml
deusebio: > I'm not sure why we only checkout on pull requests.

On pull request we don't checkout, since we want to use the tip of the the PR to indeed tests the current changes. On the other hand, when we dispatch and run the action manually, on the other hand, we want to switch to the track provided as input in the dispatch. 

is this clearer?

modules/kubeflow-mlflow/tests/test_deployment.py
deusebio: Uhm, interesting. I'm not really sure how this got it actually. I'll remove this

NohaIhab (commented):
thanks @deusebio, +1 to include #82 in the changes once it is merged. Can you also link a test run for the workflow dispatch with a different risk input?

deusebio:
> Can you also link a test run for the workflow dispatch with a different risk input?

That's really a great idea! I always forget that - as long as the CI is already in main - you can already test the dispatch flow! :D 
Let me give it a go [here](https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/18491872746/job/52687185432).

Ideally, I believe we would also like to test to change the track (e.g. using 1.9), but I'm afraid that the 1.9 track does not yet support using the risk right now. So it would probably fail. I would honestly approach this in a different ticket

deusebio:
@NohaIhab it didn't work unfortunately because we don't yet have the risk in the main branch. But this made me realize that probably it is best to NOT use the "CKF version" (as in the github issue), but just the branch, like `track/1.10`, this way we can also run tests on dev branches. 

What do you think folks?

deusebio:
Tested on dispatch [here](https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/18536236908)

mvlassis (approved):
Makes more sense to also allow non `track/*` branches :)

NohaIhab (approved):

deusebio:
I have tested that the workflow inputs work as expected:
* [AKS](https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/18723318054)
* [EKS](https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/18723439674)

The tests shows that tests can be triggered with the input provided appropriately. However, there is a *regular* failure on the EKS cluster that will need to be investigated separately, issue #87

NohaIhab (approved):

mvlassis (approved):
Thanks for raising a relevant issue @deusebio!

PR #82, opened 2025-10-10
[KF-7944] Enabling MicroK8s testing

deusebio:
Addressing #67 using MicroK8s. 

Using Canonical K8s is a bit more tricky, as discussed in the issue


.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -0,0 +1,122 @@
...
+          pip install tox
+          pip install poetry
+
+      - name: Setup operator environment
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+          juju-channel: "3.6/stable"
+          provider: microk8s
+          channel: ${{ format('{0}-strict/stable', env.K8S_VERSION) }}
+          bootstrap-options: "--agent-version 3.6.9"
```
NohaIhab: question: why pin the agent version? I think we should constantly test with the latest agent.

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -0,0 +1,122 @@
...
+          microk8s-addons: "rbac dns minio metallb:10.64.140.43-10.64.140.49"
+
+      - name: Setup terraform
+        run: |          
+          sudo snap install terraform  --channel latest/stable --classic
+          terraform --version
+
+      - name: Deploy solution
+        run: |
+          # juju switch concierge-k8s
```
NohaIhab: nit: remove comment

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -0,0 +1,122 @@
...
+      matrix:
+        bundle:
+          - module: kubeflow-mlflow
+            uats:
+              env: uats-remote
+              extra-args: --filter "not feast"
+          - module: kubeflow-feast
+            uats:
+              env: feast-remote
+              extra-args: ""
```
NohaIhab, line 31: suggestion(non-blocking): would be great to have a check on the solution modified and only run the tests for that solution (`kubeflow-mlflow` or `kubeflow-feast`), the scheduled one should be for both however. I would not block this PR on it, it's an enhancement we can do in a follow-up.

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -0,0 +1,122 @@
+name: Validation on MicroK8s
+on:
+  workflow_dispatch:
+    inputs:
+      k8s_version:
+        description: 'Kubernetes version to be used for the AKS cluster. Make sure that the corresponding K8s version is supported by the cloud.'
```
NohaIhab:
```suggestion
        description: 'Kubernetes version to be used for Microk8s.'
```

NohaIhab (commented):
thanks @deusebio, left some comments

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -0,0 +1,122 @@
...
+        name: Setup Python
+        uses: actions/setup-python@v5.0.0
+        with:
+          python-version: "3.12"
+          architecture: x64
+
+      - name: Install tox & poetry
+        run: |
+          pip install tox
+          pip install poetry
```
mvlassis, line 62: Suggestion: Let's use `pipx` to install both packages, as described [here](https://github.com/canonical/bundle-kubeflow/issues/1289)
```suggestion
          pipx install tox
          pipx install poetry
```

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -0,0 +1,122 @@
...
+
+      - name: Setup operator environment
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+          juju-channel: "3.6/stable"
```
mvlassis, line 67: suggestion: We should also add a specific charmcraft channel
```suggestion
          juju-channel: "3.6/stable"
          charmcraft-channel: "latest/stable"
```

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -0,0 +1,122 @@
...
+          provider: microk8s
+          channel: ${{ format('{0}-strict/stable', env.K8S_VERSION) }}
+          bootstrap-options: "--agent-version 3.6.9"
+          microk8s-group: snap_microk8s
+          microk8s-addons: "rbac dns minio metallb:10.64.140.43-10.64.140.49"
```
mvlassis: suggestion(blocking): I don't think we actually need the `minio` add-on here, which just deploys `minio-operator` as described in https://microk8s.io/docs/addon-minio. When installing via microk8s (which we currently do in our public documentation), we run `sudo microk8s enable dns hostpath-storage metallb:10.64.140.43-10.64.140.49 rbac`
```suggestion
          microk8s-addons: "rbac dns hostpath-storage metallb:10.64.140.43-10.64.140.49"
```

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -0,0 +1,122 @@
...
+          echo "UATS_BRANCH=${UATS_BRANCH}" >> $GITHUB_ENV
+
+      - uses: actions/checkout@v4
+
+      - id: setup-python
+        name: Setup Python
+        uses: actions/setup-python@v5.0.0
+        with:
+          python-version: "3.12"
+          architecture: x64
```
mvlassis: nitpick: The `architecture` input here is probably not needed, since by default the action uses the host OS architecture: https://github.com/actions/setup-python?tab=readme-ov-file#supported-architectures
```suggestion
```

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -0,0 +1,122 @@
...
+          done
+
+      - name: Update ENV variables from inputs if available
+        run: |
+          K8S_VERSION=${{ inputs.k8s_version || env.K8S_VERSION }}
+          echo "K8S_VERSION=${K8S_VERSION}" >> $GITHUB_ENV
+          UATS_BRANCH=${{ inputs.uats_branch || env.UATS_BRANCH }}
+          echo "UATS_BRANCH=${UATS_BRANCH}" >> $GITHUB_ENV
+
+      - uses: actions/checkout@v4
```
mvlassis: nitpick: Add name and use latest version as we do in other repositories
```suggestion
    - name: Check out code
      uses: actions/checkout@v5
```

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -0,0 +1,122 @@
...
+    runs-on: ["self-hosted-linux-amd64-jammy-xlarge"]
+
+    strategy:
+      fail-fast: false
+      matrix:
+        bundle:
+          - module: kubeflow-mlflow
+            uats:
+              env: uats-remote
+              extra-args: --filter "not feast"
```
mvlassis, line 27: suggestion(blocking): We *should* also test `kubeflow` on its own, right? This would help us more easily identify issues with the `mlflow` integration
```suggestion
          - module: kubeflow
            uats:
              env: kubeflow-remote
          - module: kubeflow-mlflow
            uats:
              env: uats-remote
              extra-args: --filter "not feast"
```

mvlassis (commented):
Left some small comments

.github/workflows/deploy-to-microk8s.yaml
deusebio: Yes, we have been pinning this as `3.6.10` was rather flaky some weeks ago and we pinned this in Spark (where I also copy/pasted part of the code). But I can give it a go to remove it :+1:

.github/workflows/deploy-to-microk8s.yaml
deusebio: I would *personally* be against this, and I'd rather be safer than sorrier. I mean, I would generally always go ahead to tests more, even if it is not needed (for instance, if we change common modules - e.g. kubeflow - we should tests both, if we change submodules only some), since sometimes there may also be changes that are subtle sneaking into. Similarly, we generally don't run tests (e.g. integrations) based on the changes that we make also in charms, so I'm not sure why we should here. On the other hand, putting logic to avoid some tests based on the changes would expose us to the risk of having bugs/regressions sneaking in, I feel. 

Sometimes we can restrict the testing on PRs (e.g. in Spark we had 100 CIs that we then subfiltered in a reasonable manner to cover most of the use-cases), but I would do this based on the need. One CI instead of two does not seem to provide enough benefits or justification to limit the testing IMHO

.github/workflows/deploy-to-microk8s.yaml
deusebio: We don't have test module for this, compare `modules/kubeflow` with `modules/kubeflow-mlflow`. This is not supported at the moment, and I would not add this as part of the current PR

.github/workflows/deploy-to-microk8s.yaml
deusebio: Done!

.github/workflows/deploy-to-microk8s.yaml
deusebio: Done! i have used `3.x/stable` as I really don't like to use `latest/stable`

.github/workflows/deploy-to-microk8s.yaml
deusebio: Done!

.github/workflows/deploy-to-microk8s.yaml
deusebio: Done!

.github/workflows/deploy-to-microk8s.yaml
deusebio: done!

deusebio:
@mvlassis I gave it a go to follow up on your comments, but getting it to work does not seem straight forward:

1. When using `noble` the MicroK8s setup hangs on the MetalLB setup (see [here](https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/18568015537/job/52934313467) and [here](https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/18568460577/job/52935883226))

2. Using `pipx` (installing it as advised [here](https://github.com/canonical/bundle-kubeflow/issues/1289)) on `jammy` would fall back to use the default `3.10` python version which is not compatible with the UATs (that requires 3.12) ([reference](https://github.com/canonical/charmed-kubeflow-solutions/actions/runs/18565813419/job/52926490169))

Fixing this may require a bit more effort and investigation, but I *personally* don't think it is worth it given MicroK8s are still some placeholder to be replaced by Canonical K8s tests, where it is probably more critical to have a crisp and up to date action

.github/workflows/deploy-to-microk8s.yaml
mvlassis: +1 for Enrico here, let's test all our modules everytime. The tests run in parallel, so we wouldn't save time either way.

.github/workflows/deploy-to-microk8s.yaml
mvlassis: Created https://github.com/canonical/charmed-kubeflow-solutions/issues/86 to track this

mvlassis:
We could investigate issues `1.` to `2.`, but I'm OK with approving and spending more time to enable testing for Canonical K8s which is more important in our case

mvlassis (approved):

NohaIhab (approved):
approving to merge the replacement for AKS, we should look into failures or rather work on using Canonical k8s sooner rather than later.

deusebio:
@NohaIhab It had to re-instantiate the juju agent pinning to get the tests to pass. Given the stability issues, I would honestly nudge to use pinning and then update the pinning from time to time with a self-contained PR. This way it becomes easier to have reproducible envs, pin-point regressions, and poke the Juju team accordingly. Doing this within feature work makes things a bit tricky honestly

PR #91, opened 2025-11-19
Expose config option for `kubeflow_profiles` security policy

### LINKED ISSUE #90: Exposing Configs for Applying `baseline` Pod Security Standards to User Namespaces
### Context

Refer to https://github.com/canonical/kubeflow-profiles-operator/issues/269 for context.

### What needs to get done

- expose the newly created charm config option from the Profile controller to the solutions' modules

### Definition of Done

- the newly created charm config option is exposed from the Profile controller to the solutions' modules

mvlassis:
Closes #90 

This PR exposes the `security-policy` config option of `kubeflow-profiles`, that has been created in https://github.com/canonical/kubeflow-profiles-operator/pull/270.

### DIFF
--- a/.github/workflows/deploy-to-microk8s.yaml
+++ b/.github/workflows/deploy-to-microk8s.yaml
@@ -8,6 +8,7 @@ on:
       uats_branch:
         description: 'Branch to run the UATs from e.g. main or track/1.10. By default, this is defined by the dependencies.yaml file.'
         required: false
+        default: "track/1.10"
       ckf_branch:
         description: 'Branch on the Charmed Kubeflow repository to be tested'
         required: false
@@ -77,7 +78,7 @@ jobs:
           juju-channel: "3.6/stable"
           charmcraft-channel: "3.x/stable"
           provider: microk8s
-          channel: ${{ format('{0}-strict/stable', env.K8S_VERSION) }}
+          channel: "1.32-strict/stable"
           microk8s-group: snap_microk8s
           microk8s-addons: "rbac dns hostpath-storage metallb:10.64.140.43-10.64.140.49"
           bootstrap-options: "--agent-version 3.6.9"
@@ -100,7 +101,7 @@ jobs:
           git clone https://github.com/canonical/charmed-kubeflow-uats.git ~/charmed-kubeflow-uats
           cd ~/charmed-kubeflow-uats
           git checkout ${{ env.UATS_BRANCH }}
-          tox -e ${{ matrix.bundle.uats.env }} -- ${{ matrix.bundle.uats.extra-args }}
+          tox -e ${{ matrix.bundle.uats.env }} -- --bundle "" ${{ matrix.bundle.uats.extra-args }}
 
       # On failure, capture debugging resources
       - name: Select model
--- a/modules/kubeflow-feast/variables.tf
+++ b/modules/kubeflow-feast/variables.tf
@@ -65,6 +65,18 @@ variable "https_proxy" {
   default     = ""
 }
 
+variable "istio_cni_bin_dir" {
+  description = "Path to CNI binaries, e.g. /opt/cni/bin. If not provided, the Istio control plane will be installed/upgraded with the Istio CNI plugin disabled. This path depends on the Kubernetes installation, please refer to https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/ for information to find out the correct path."
+  type        = string
+  default     = ""
+}
+
+variable "istio_cni_conf_dir" {
+  description = "Path to conflist files describing the CNI configuration, e.g. /etc/cni/net.d. If not provided, the Istio control plane will be installed/upgraded with the Istio CNI plugin disabled. This path depends on the Kubernetes installation, please refer to https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/ for information to find out the correct path."
+  type        = string
+  default     = ""
+}
+
 variable "istio_tls_secret_id" {
   description = "The juju secret id for the tls key/cert for istio-pilot"
   type        = string
@@ -87,6 +99,17 @@ variable "kfp_db_size" {
   description = "KFP database storage size"
   type        = string
   default     = "10G"
+}
+
+variable "kubeflow_profiles_security_policy" {
+  description = "Security policy for pod security standards enforced in user workloads. Only `privileged` and `baseline` are supported"
+  type        = string
+  default     = "privileged"
+
+  validation {
+    condition     = var.kubeflow_profiles_security_policy == "privileged" || var.kubeflow_profiles_security_policy == "baseline"
+    error_message = "The variable must be one of 'privileged' or 'baseline'."
+  }
 }
 
 variable "minio_size" {
--- a/modules/kubeflow/applications.tf
+++ b/modules/kubeflow/applications.tf
@@ -240,7 +240,7 @@ module "kubeflow_dashboard" {
 module "kubeflow_profiles" {
   source     = "git::https://github.com/canonical/kubeflow-profiles-operator//terraform?ref=track/1.10"
   model_name = var.create_model ? juju_model.kubeflow[0].name : local.model
-  config {
+  config = {
 	"security-policy" : var.kubeflow_profiles_security_policy
   }
   revision   = var.kubeflow_profiles_revision
--- a/modules/kubeflow/applications.tf
+++ b/modules/kubeflow/applications.tf
@@ -238,7 +238,7 @@ module "kubeflow_dashboard" {
 }
 
 module "kubeflow_profiles" {
-  source     = "git::https://github.com/canonical/kubeflow-profiles-operator//terraform?ref=main"
+  source     = "git::https://github.com/canonical/kubeflow-profiles-operator//terraform?ref=track/1.10"
   model_name = var.create_model ? juju_model.kubeflow[0].name : local.model
   config = {
     "security-policy" : var.kubeflow_profiles_security_policy
--- a/modules/kubeflow/applications.tf
+++ b/modules/kubeflow/applications.tf
@@ -53,6 +53,8 @@ module "istio_pilot" {
   model_name = var.create_model ? juju_model.kubeflow[0].name : local.model
   config = {
     default-gateway = "kubeflow-gateway",
+    "cni-bin-dir" : var.istio_cni_bin_dir
+    "cni-conf-dir" : var.istio_cni_conf_dir
     "tls-secret-id" : var.istio_tls_secret_id
   }
   revision = var.istio_pilot_revision
--- a/modules/kubeflow/applications.tf
+++ b/modules/kubeflow/applications.tf
@@ -240,6 +240,9 @@ module "kubeflow_dashboard" {
 module "kubeflow_profiles" {
   source     = "git::https://github.com/canonical/kubeflow-profiles-operator//terraform?ref=track/1.10"
   model_name = var.create_model ? juju_model.kubeflow[0].name : local.model
+  config {
+	"security-policy" : var.kubeflow_profiles_security_policy
+  }
   revision   = var.kubeflow_profiles_revision
   channel    = "1.10/${var.risk}"
 }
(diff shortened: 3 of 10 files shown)


modules/kubeflow/variables.tf
```diff
@@ -101,6 +101,12 @@ variable "kfp_db_size" {
   default     = "10G"
 }
 
+variable "kubeflow_profiles_security_policy" {
+  description = "Pod security policy to enforce in user workloads"
```
MattiaSarti: suggestion (blocking): actually, "pod security policy" would refer to [an old, deprecated K8s feature](https://kubernetes.io/docs/concepts/security/pod-security-policy/), so I would suggest describing the option as **_"Security policy for pod security standards enforced in user workloads - only `privileged` and `baseline` supported"_** (hinting also to the fact `restricted` is not supported yet)

modules/kubeflow/variables.tf
```diff
@@ -101,6 +101,12 @@ variable "kfp_db_size" {
   default     = "10G"
 }
 
+variable "kubeflow_profiles_security_policy" {
```
MattiaSarti, line 116: suggestion (blocking): shall we document this new input in the README as well?

modules/kubeflow/variables.tf
MattiaSarti, line 1: todo: sorry to ask you even if slightly unrelated to this pull request, but could you please also expose the options for [Istio CNI](https://github.com/canonical/bundle-kubeflow/issues/1356#issuecomment-3516535477)? I've just realized it would be way more convenient for users to have them as inputs to more easily enable [this change](https://documentation.ubuntu.com/charmed-kubeflow/how-to/manage/enable-istio-cni/#enable-the-plugin), required for `baseline` pod security standards

MattiaSarti (changes requested):

modules/kubeflow/variables.tf
```diff
@@ -101,6 +101,12 @@ variable "kfp_db_size" {
...
+  description = "Pod security policy to enforce in user workloads"
+  type        = string
+  default     = "privileged"
+}
```
MattiaSarti, line 125: suggestion (blocking): can we add some validation for the only two acceptable input values, like [here](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.10/modules/kubeflow/variables.tf#L6-L9)?

MattiaSarti (changes requested):

modules/kubeflow/variables.tf
mvlassis: Updated in https://github.com/canonical/charmed-kubeflow-solutions/pull/91/commits/668d077230ab4f900ffb9238f1e2a3b0d0bfcc1a

modules/kubeflow/variables.tf
mvlassis: Updated in https://github.com/canonical/charmed-kubeflow-solutions/pull/91/commits/668d077230ab4f900ffb9238f1e2a3b0d0bfcc1a

modules/kubeflow/variables.tf
mvlassis: Added validation in https://github.com/canonical/charmed-kubeflow-solutions/pull/91/commits/668d077230ab4f900ffb9238f1e2a3b0d0bfcc1a (note that `contains` is more permissive than needed, so I did exact matching)

MattiaSarti: Wonderful idea!

modules/kubeflow/variables.tf
mvlassis: Exposed the 2 options in https://github.com/canonical/charmed-kubeflow-solutions/pull/91/commits/ee2bce24b14e4061851ee28bcf8dfae051475239

MattiaSarti (changes requested):

MattiaSarti (approved):

MattiaSarti:
Let's not merge yet, though.

MattiaSarti:
@mvlassis, we forgot to expose the same configs to the modules for MLFlow and Feast

MattiaSarti (approved):

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -8,6 +8,7 @@ on:
       uats_branch:
         description: 'Branch to run the UATs from e.g. main or track/1.10. By default, this is defined by the dependencies.yaml file.'
         required: false
+        default: "track/1.10"
```
deusebio, line 11: **issue** Why are we providing the defaults here? The defaults are bundled in the `dependencies.yaml` file by design, so we should rather add the default there. But the current business logic, tries to make it as consistent as possible, so there is auto-inference if this is not provided [here](https://github.com/canonical/charmed-kubeflow-solutions/pull/91/files#diff-71cc896038d3c115e59bb1b36fbd1c8d9e1ab9fc160043b62a5d605d603d79b3R61), that sets this consistently with the branch of CKF being tested. 

I realize that the description may be a bit misleading and provided a suggestion above.

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -8,6 +8,7 @@ on:
       uats_branch:
         description: 'Branch to run the UATs from e.g. main or track/1.10. By default, this is defined by the dependencies.yaml file.'
```
deusebio, line 9:
```suggestion
        description: 'Branch to run the UATs from e.g. main or track/1.10. By default, this falls back to the name of the CKF branch being tested.'
```

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -77,7 +78,7 @@ jobs:
           juju-channel: "3.6/stable"
           charmcraft-channel: "3.x/stable"
           provider: microk8s
-          channel: ${{ format('{0}-strict/stable', env.K8S_VERSION) }}
+          channel: "1.32-strict/stable"
```
deusebio, line 81: **issue** why are we doing this? This should not be needed, since the K8S_VERSION is the one provided in the `dependencies.yaml`, but if we hard code this, testing other K8s versions (even with manual trigger would not work), and also the fact that we expose the input option would provide wrong expectation that we can

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -100,7 +101,7 @@ jobs:
           git clone https://github.com/canonical/charmed-kubeflow-uats.git ~/charmed-kubeflow-uats
           cd ~/charmed-kubeflow-uats
           git checkout ${{ env.UATS_BRANCH }}
-          tox -e ${{ matrix.bundle.uats.env }} -- ${{ matrix.bundle.uats.extra-args }}
+          tox -e ${{ matrix.bundle.uats.env }} -- --bundle "" ${{ matrix.bundle.uats.extra-args }}
```
deusebio, line 104: **todo** this should be removed before merging

modules/kubeflow-feast/variables.tf
```diff
@@ -65,6 +65,18 @@ variable "https_proxy" {
   default     = ""
 }
 
+variable "istio_cni_bin_dir" {
+  description = "Path to CNI binaries, e.g. /opt/cni/bin. If not provided, the Istio control plane will be installed/upgraded with the Istio CNI plugin disabled. This path depends on the Kubernetes installation, please refer to https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/ for information to find out the correct path."
+  type        = string
+  default     = ""
```
deusebio, line 71: **question** shouldn't this be null? Is providing null or "" the same?

modules/kubeflow-feast/variables.tf
```diff
@@ -89,6 +101,17 @@ variable "kfp_db_size" {
   default     = "10G"
 }
 
+variable "kubeflow_profiles_security_policy" {
+  description = "Security policy for pod security standards enforced in user workloads. Only `privileged` and `baseline` are supported"
+  type        = string
+  default     = "privileged"
+
+  validation {
+    condition     = var.kubeflow_profiles_security_policy == "privileged" || var.kubeflow_profiles_security_policy == "baseline"
```
deusebio, line 110: **niptick,suggestion** A slightly more scalable implementation would be:

```
validation {
    condition     = contains(["baseline", "privileged"], var.kubeflow_profiles_security_policy)
    error_message = "Valid values for var.kubeflow_profiles_security_policy are (baseline, privileged)"
  }
```

Should we one day also support "restricted"

modules/kubeflow/applications.tf
```diff
@@ -240,8 +242,11 @@ module "kubeflow_dashboard" {
 module "kubeflow_profiles" {
   source     = "git::https://github.com/canonical/kubeflow-profiles-operator//terraform?ref=track/1.10"
   model_name = var.create_model ? juju_model.kubeflow[0].name : local.model
-  revision   = var.kubeflow_profiles_revision
-  channel    = "1.10/${var.risk}"
+  config = {
+    "security-policy" : var.kubeflow_profiles_security_policy
+  }
+  revision = var.kubeflow_profiles_revision
+  channel  = "latest/edge"
```
deusebio, line 249: **todo** let's not forget to fix this. As the feature landed already on 1.10/edge? If so, (**suggestion**) we could have JUST used the risk to avoid having this change to cripple in the release branch

deusebio (changes requested):
The config variable additions and terraform modifications are fine, but there are a few points that needs to be fixed before merging this to make sure we don't break the release branch

deusebio:
By the way, it is probably best to change the base to a feature branch, such that this and the integation tests in https://github.com/canonical/charmed-kubeflow-solutions/pull/94 can be merged into the stable branch together

modules/kubeflow-feast/variables.tf
MattiaSarti: Not exactly, I am linking [this](https://github.com/canonical/charmed-kubeflow-solutions/pull/94#discussion_r2576432137) for context.

MattiaSarti:
> By the way, it is probably best to change the base to a feature branch, such that this and the integation tests in https://github.com/canonical/charmed-kubeflow-solutions/pull/94 can be merged into the stable branch together

@deusebio, is it fine if we merge https://github.com/canonical/charmed-kubeflow-solutions/pull/94 here after its comments are closed and then test everything together in this same PR?

modules/kubeflow/applications.tf
MattiaSarti: Of course, we'll restore it before merging. @mvlassis, let's not resolve this comment in order not to forget.

But it's not there yet, @deusebio, as we only merged the changes to `main` for that charm and it's only available on channel `latest/edge` so far.

@mvlassis, the time has probably come to backport it to branch `track/1.10` to have it on channel `1.10/edge`.

.github/workflows/deploy-to-microk8s.yaml
MattiaSarti: @mvlassis, let's not resolve this comment in order not to forget.

.github/workflows/deploy-to-microk8s.yaml
MattiaSarti: I am not sure why this was necessary, but if @mvlassis can undo this change then we can merge https://github.com/canonical/charmed-kubeflow-solutions/pull/94 into this branch and have that PR solve the parametrization, as it's already addressing this.

deusebio:
> is it fine if we merge https://github.com/canonical/charmed-kubeflow-solutions/pull/94 here after its comments are closed and then test everything together in this same PR?

I'd rather use feature branches honestly, given the code comes from two different authors and also it can be logically separated:
1. feature work by adding config options
2. Test/CI work 

That's what I would do, but up to you, if you prefer to merge the PRs together that's fine, but there should be only ONE driver of the the PR then. 

On feature branch you can also just review among the two of you. The final PR should then have the 2 reviewers constraints, as usual

MattiaSarti:
Thanks, @deusebio, it makes sense. My only concern was that, if we did so, we would end up having the feature of exposing the new configs (the feature branch) merged into the branch of track 1.10 (the stable branch) before its changes could be tested with the testing logic (the test branch). Do we prefer to be cautious and merge this two branches together first (yet on a branch, be this or a new one) and validate that everything works before merging anything to track 1.10, where users could be exposed to untested problems?

_Edit: perhaps you already meant to do the same as I suggest, now that I read https://github.com/canonical/charmed-kubeflow-solutions/pull/94#discussion_r2576775444_

MattiaSarti:
Ported to https://github.com/canonical/charmed-kubeflow-solutions/pull/96

PR #93, opened 2025-11-27
[MISC] add sysctl specification

deusebio:
Adding sysctl specification as advised in the doumentation.

### DIFF
--- a/.github/workflows/deploy-to-k8s.yaml
+++ b/.github/workflows/deploy-to-k8s.yaml
@@ -38,6 +38,11 @@ jobs:
               extra-args: ""
 
     steps:
+      - name: Setup tmate session
+        uses: canonical/action-tmate@main
+        with:
+          detached: true
+
       - name: Checkout repository
         uses: actions/checkout@v5
 
--- a/.github/workflows/deploy-to-k8s.yaml
+++ b/.github/workflows/deploy-to-k8s.yaml
@@ -79,16 +79,16 @@ jobs:
       - name: Setup environment
         run: |
           # Avoiding dockerhub rate limits (see https://canonical-self-hosted-github-runner-docs.readthedocs-hosted.com/en/latest/usage/faq/how-to-avoid-dockerhub-rate-limits/ for more informations)
-          # if [ -n "$DOCKERHUB_MIRROR" ]; then
-          # MIRROR_CONFIG=/etc/containerd/hosts.d/docker.io
-          #   
-          # sudo mkdir -p ${MIRROR_CONFIG}
-          # sudo chown $USER ${MIRROR_CONFIG}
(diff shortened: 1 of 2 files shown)


MattiaSarti (approved):

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -38,6 +38,11 @@ jobs:
               extra-args: ""
 
     steps:
+      # - name: Setup tmate session
+      #   uses: canonical/action-tmate@main
+      #   with:
+      #     detached: true
```
mvlassis: nitpick: Let's remove unused code
```suggestion
```

deusebio: I generally find it useful, since I always end up googleing this, but I understand the point, so I've accepted the suggestion

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -96,6 +97,11 @@ jobs:
           sudo concierge prepare --trace
           cd ..
 
+      - name: Configure sysctl
+        run: |
+          sudo sysctl fs.inotify.max_user_instances=1280
+          sudo sysctl fs.inotify.max_user_watches=655360
```
mvlassis, line 103: This part shouldn't be necessary here, the `k8s` snap already modifies the system config [here](https://github.com/canonical/k8s-snap/blob/64cee640bc1abcded5a3c6070cc6ed40b74ee2a8/src/k8s/pkg/snap/snap.go#L240-L245) to set:
```
fs.inotify.max_user_instances = 1024
fs.inotify.max_user_watches = 1048576
```
```suggestion
```

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -82,6 +82,11 @@ jobs:
           microk8s-addons: "rbac dns hostpath-storage metallb:10.64.140.43-10.64.140.49"
           bootstrap-options: "--agent-version 3.6.9"
 
+      - name: Configure sysctl
+        run: |
+          sudo sysctl fs.inotify.max_user_instances=1280
+          sudo sysctl fs.inotify.max_user_watches=655360
+
```
mvlassis, line 89: The same note as above applies to `microk8s` as well, as by default the `inotify` limits are set to high values:
```
fs.inotify.max_user_instances = 1024
fs.inotify.max_user_watches = 1048576
```
A person can check this by uninstalling microk8s, setting a low value like `sudo sysctl fs.inotify.max_user_instances=128;
sudo sysctl fs.inotify.max_user_watches=8192`, and reinstalling `microk8s`

```suggestion
```

deusebio:
Let's close the PR then. I have raise a PR to update the documentation https://github.com/canonical/charmed-kubeflow-documentation/pull/28

PR #94, opened 2025-11-28
ci: test on MicroK8s both with `baseline` and `privileged` Pod Security Standards

MattiaSarti:
This pull request resolves https://github.com/canonical/charmed-kubeflow-solutions/issues/95 and:
- it is raised to another branch in order not to modify the official solutions for track 1.10 until all changes are tested and approved
- it relies on a branch of the UATs to co-test both changes before merging anything :warning: this reference will be reverted once the UATs' PR is merged to the branch for track 1.10


.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -38,6 +38,13 @@ jobs:
             uats:
               env: feast-remote
               extra-args: ""
+        pod-security-standards:
```
deusebio, line 41: **praise** This is actually very good that we tests both istio CNI and normal side-car init container injection. Beside also testing the PSS (where the istio CNI is required) we also have coverage for both istio modes actually, as a by-product result

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -94,14 +124,14 @@ jobs:
...
-          tox -c ./modules/${{ matrix.bundle.module }} -vve test_deployment -- -vv -s --risk ${{ env.RISK }}
+          tox -c ./modules/${{ matrix.bundle.module }} -vve test_deployment -- -vv -s --risk ${{ env.RISK }} --istio-cni-bin-dir ${{ matrix.pod-security-standards.istio-cni-bin-dir }} --istio-cni-conf-dir ${{ matrix.pod-security-standards.istio-cni-conf-dir }} --pss ${{ matrix.pod-security-standards.policy }}
 
       - name: Run UATs
         run: |
           git clone https://github.com/canonical/charmed-kubeflow-uats.git ~/charmed-kubeflow-uats
           cd ~/charmed-kubeflow-uats
-          git checkout ${{ env.UATS_BRANCH }}
-          tox -e ${{ matrix.bundle.uats.env }} -- --bundle "" ${{ matrix.bundle.uats.extra-args }}
+          git checkout kf-8136/test/job-exempt-from-baseline-policy
```
deusebio: **todo** this should be changed eventually. Let's just not forget about it

(small niptick is that I would have probably changed the default, not the hard code value, but this is really minor since this change will be reverted)

modules/kubeflow-feast/tests/conftest.py
```diff
@@ -34,6 +59,21 @@ def pytest_addoption(parser):
         help="Risk to be used when deploying the terraform module",
     )
 
+@pytest.fixture(scope="module")
+def istio_cni_bin_dir(request) -> str:
```
deusebio, line 63:
```suggestion
def istio_cni_bin_dir(request) -> str | None:
```

I suppose this can be none if the option is not declared. Similarly also below

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -73,19 +80,42 @@ jobs:
...
+          # and Juju does not support non-strict MicroK8s snaps unless manually
+          # added to Juju as K8s clouds like this...
+          # `sudo microk8s config | juju add-k8s local-k8s --client`
+          # ...which neither the action `charmed-kubernetes/actions-operator`
+          # nor the `concierge` snap support configuring as of today
+
+          sudo snap refresh
+          sudo snap install juju --channel=3.6/stable --classic
+          sudo snap install charmcraft --channel=3.x/stable --classic
+          sudo snap install microk8s --channel=1.32/stable --classic
```
deusebio: **todo** the K8s version should be configurable

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -73,19 +80,42 @@ jobs:
...
+          # https://github.com/canonical/istio-operators/issues/656
+          # ...non-strict MicroK8s is required for Istio CNI to be functional,
+          # and Juju does not support non-strict MicroK8s snaps unless manually
+          # added to Juju as K8s clouds like this...
+          # `sudo microk8s config | juju add-k8s local-k8s --client`
+          # ...which neither the action `charmed-kubernetes/actions-operator`
+          # nor the `concierge` snap support configuring as of today
+
+          sudo snap refresh
+          sudo snap install juju --channel=3.6/stable --classic
```
deusebio: **todo** the Juju version should be configurable. We actually probably forgot about this (and that was probably a bug) but we should use the one provided in `dependencies.yaml` (since we are not exposing this as input in the workflow)

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -73,19 +80,42 @@ jobs:
...
+
+          microk8s status --wait-ready
+          sudo microk8s enable dns hostpath-storage rbac metallb:10.64.140.43-10.64.140.49
+          microk8s status --wait-ready
+
+          # a workaround for Juju 3's strict snap:
+          mkdir -p ~/.local/share/juju
+
+          sudo microk8s config | juju add-k8s local-k8s --client
+          juju bootstrap --agent-version 3.6.9 local-k8s microk8s
```
deusebio: **minor,nice-to-have** can we try to remove the `--agent-version`? We pinned this since 3.6.10 was not very stable, but this should now be resolved

modules/kubeflow-feast/tests/test_deployment.py
```diff
@@ -27,6 +34,12 @@ async def test_apply_terraform_solution(self, juju: jubilant.Juju, risk):
                 "-var",
                 "create_model=false",
                 "-var",
+                f"istio_cni_bin_dir={istio_cni_bin_dir}",
```
deusebio, line 37: **todo** the latest modifications of the deployments have been bundling the vars into "sections", see [here](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.10/modules/kubeflow-mlflow/tests/conftest.py#L39). It would probably be nice to have a single fixture about the "privileged" part

modules/kubeflow-feast/main.tf
```diff
@@ -1,62 +1,62 @@
...
-  grafana_agent_k8s_size           = var.grafana_agent_k8s_size
-  http_proxy                       = var.http_proxy
-  https_proxy                      = var.https_proxy
-  istio_cni_bin_dir = var.istio_cni_bin_dir
-  istio_cni_conf_dir = var.istio_cni_conf_dir
-  istio_tls_secret_id              = var.istio_tls_secret_id
-  jupyter_ui_config                = var.jupyter_ui_config
-  katib_db_size                    = var.katib_db_size
-  kfp_db_size                      = var.kfp_db_size
+  source                            = "../kubeflow"
```
deusebio, line 2: **niptick** this linting would seem to rather belong to #91

modules/kubeflow-feast/main.tf
```diff
@@ -1,62 +1,62 @@
...
-  metacontroller_operator_revision = var.metacontroller_operator_revision
-  mlmd_revision                    = var.mlmd_revision
-  minio_revision                   = var.minio_revision
-  oidc_gatekeeper_revision         = var.oidc_gatekeeper_revision
-  pvcviewer_operator_revision      = var.pvcviewer_operator_revision
-  tensorboard_controller_revision  = var.tensorboard_controller_revision
-  tensorboards_web_app_revision    = var.tensorboards_web_app_revision
-  training_operator_revision       = var.training_operator_revision
-  grafana_agent_k8s_revision       = var.grafana_agent_k8s_revision
+  minio_size                        = var.minio_size
```
deusebio, line 20: **niptick** this linting would rather seem to belong to #91

deusebio (changes requested):
Overall quite good! Just a few comments attached. Nothing major. Possibly the only one required for approval on my part would be [this](https://github.com/canonical/charmed-kubeflow-solutions/pull/94/files#r2576376032)

.github/workflows/deploy-to-microk8s.yaml
MattiaSarti: Thank you!

modules/kubeflow-feast/tests/conftest.py
MattiaSarti: Wouldn't it be better to keep them as strings and only strings, as [the underlying charm interprets only strings and wants empty ones to disable Istio CNI](https://github.com/canonical/istio-operators/blob/main/charms/istio-pilot/config.yaml#L2-L15)? It may be less risk prone to avoid the conversion:
_<ins>only</ins> strings CLI args -> <ins>either</ins> strings <ins>or</ins> Nones in Python -> <ins>only</ins> strings in Terraform (-> <ins>only</ins> strings as charm configs)_

This way we always propagate them as is from the module to the charm instead of risking not to because of Python logic bases on `None`s.

modules/kubeflow-feast/main.tf
MattiaSarti: Yes, but we are planning to merge this PR into that, as it starts from the tip of that branch, and so this will eventually fix it there too.

modules/kubeflow-feast/main.tf
MattiaSarti: Yes, but we are planning to merge this PR into that, as it starts from the tip of that branch, and so this will eventually fix it there too.

modules/kubeflow-feast/tests/test_deployment.py
MattiaSarti: Good idea! The problem is that we should rebase to achieve that, as those changes were not on main when we started the PR this PR is built onto (https://github.com/canonical/charmed-kubeflow-solutions/pull/91) and then this PR itself (https://github.com/canonical/charmed-kubeflow-solutions/pull/94). Can we fix this on https://github.com/canonical/charmed-kubeflow-solutions/pull/91 once we merge https://github.com/canonical/charmed-kubeflow-solutions/pull/94 into https://github.com/canonical/charmed-kubeflow-solutions/pull/91, so that it's easier to rebase?

modules/kubeflow-feast/tests/conftest.py
deusebio: As it is written, I believe the type *could* also be None. To make it type compatible just use:

```
return request.config.getoption("--istio-cni-bin-dir", "")
```

and you could also use skip=False to raise

See [here](https://github.com/pytest-dev/pytest/blob/8fb7815f1440c7cf8718c248639109a779db575d/src/_pytest/config/__init__.py#L1877)

modules/kubeflow-feast/tests/test_deployment.py
deusebio: I *personally* believe we should use feature branches, with #91 being merged into the feature branch, and then 94 into the feature branch.

But my understanding is that #91 is about providing the config options (leaving outside the CI/tests), whereas this one is about the test. As such, these changes would rather belong to the tests and CI, I believe

deusebio: anyhow, up to you on how to proceed. If you want to merge anyway and work on the comments onthe other one that's also fine, see https://github.com/canonical/charmed-kubeflow-solutions/pull/91#issuecomment-3596152354

In the latter case, my approval is not required, I'd say

modules/kubeflow-feast/tests/conftest.py
MattiaSarti: Agreed with call: `request.config.getoption("--istio-cni-bin-dir") or ""`

MattiaSarti:
Ported to https://github.com/canonical/charmed-kubeflow-solutions/pull/96

PR #92, opened 2025-11-27
[KF-8177] Use track 1.10/ for minio

deusebio:
The charm on the 1.10 track implements side-car and ops pattern, and it is able to support back up and restore functionalities
 
This requires the PR [here](https://github.com/canonical/charmed-mlflow-solutions/pull/25) to be merged first

PR #96, opened 2025-12-02
feat: enable Pod Security Standards policy in Terraform modules

### LINKED ISSUE #90: Exposing Configs for Applying `baseline` Pod Security Standards to User Namespaces

mvlassis:
This PR is a merge of #91 and #94 and closes #90

## Notes
- This PR adds 3 new variables in the Terraform solutions: `istio_cni_bin_dir`, `istio_cni_conf_dir`, and `kubeflow_profiles_security_policy`. `kubeflow_profiles_security_policy` has simple validation to ensure it has appropriate values.
- These 3 variables can be configured with options when running the tests
- `deploy-to-microk8s.yaml` is modified to test the security policy in the CI
- We are adding a new value in `dependencies.yaml` to decouple the CKF branch from the UATs branch which is very useful in testing.

## Note for reviewers
There are 3 changes that will be reverted after the PR is approved so that the tests pass:
- `UATS_BRANCH` from `main` (will revert to `track/1.10`)
- `--bundle ""` in `deploy-to-microk8s.yaml` (will revert to not passing anything)
- The channel of `kubeflow-profiles` from `1.10/edge` (will revert to `1.10/stable`)


MattiaSarti (changes requested):
suggestion: Merging https://github.com/canonical/charmed-kubeflow-solutions/pull/98 should solve the missing details!

note: My PR has "TODO"s beside the lines with temporary changes to be reverted before merging, to enable temporary testing.

MattiaSarti (approved):
todo: before merging, let's undo the temporary changes for testing, the ones marked with "TODO" comments

MattiaSarti (approved):

MattiaSarti:
@deusebio and @kimwnasptd, we're requesting reviews from both of you as I've co-authored the changes and my review doesn't count.

.github/dependencies.yaml
```diff
@@ -1,3 +1,4 @@
-CKF_BRANCH: "track/1.10"
+CKF_BRANCH: "track/1.10"  
 K8S_VERSION: "1.32"
-JUJU_VERSION: "3.6"
\ No newline at end of file
+JUJU_VERSION: "3.6"
+UATS_BRANCH: "main" # TODO: revert this to "track/1.10"
```
deusebio: **question** I still don't understand why we need this. In general, we would expect that this is aligned with CKF_BRANCH?

modules/kubeflow/variables.tf
```diff
@@ -101,6 +113,16 @@ variable "kfp_db_size" {
   default     = "10G"
 }
 
+variable "kubeflow_profiles_security_policy" {
+  description = "Security policy for pod security standards enforced in user workloads. Only `privileged` and `baseline` are supported"
+  type        = string
+  default     = "privileged"
+
+  validation {
+    condition     = var.kubeflow_profiles_security_policy == "privileged" || var.kubeflow_profiles_security_policy == "baseline"
```
deusebio: **suggestion** using `contains` would seem a bit more readable and scalable, but up to you. I would just keep it consistent with the other line [here](https://github.com/canonical/charmed-kubeflow-solutions/pull/96/files#diff-05a35bba616d71ffe22a96f5106240e8e179dca22122b69f94f87c32b48d6821R128)

modules/kubeflow/applications.tf
```diff
@@ -240,8 +242,11 @@ module "kubeflow_dashboard" {
 module "kubeflow_profiles" {
   source     = "git::https://github.com/canonical/kubeflow-profiles-operator//terraform?ref=track/1.10"
   model_name = var.create_model ? juju_model.kubeflow[0].name : local.model
-  revision   = var.kubeflow_profiles_revision
-  channel    = "1.10/${var.risk}"
+  config = {
+    "security-policy" : var.kubeflow_profiles_security_policy
+  }
+  revision = var.kubeflow_profiles_revision
+  channel  = "1.10/edge" # TODO: revert the risk from edge to var.risk
```
deusebio: **todo** update

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -15,13 +15,14 @@ on:
...
         default: 'stable'
+
   pull_request:
   schedule:
     - cron: "17 02 * * 1"
 
 jobs:
   deploy-solution:
-    name: Deploy to MicroK8s
+    name: Deploy CKF + MLFlow/Feast solution to K8s
```
deusebio: **todo, minor** this would not be consistent with the other substrates. Also I would not make a title too long, "Deploy to <substrate>". Listing the subcomonents may not scale out. The names of the components are anyway defined in the name of the jobs in the matrix

deusebio (commented):
There are still few "TODO" to be fixed, and also I have a few points I'd like to follow up

.github/dependencies.yaml
MattiaSarti: On track branches, yes, but not necessarily in general: in this PR, for example, we want to test the changes of this branch of the solutions with the UATs currently on main, and we need to be able to disentangle the branches in these cases.

modules/kubeflow/variables.tf
mvlassis: Updated in https://github.com/canonical/charmed-kubeflow-solutions/pull/96/commits/fc6bec157ed049a1ef1a1f4acc073eb6ca1587dd

.github/dependencies.yaml
mvlassis: The reason is that we haven't yet backported the new UAT to [track/1.10](https://github.com/canonical/charmed-kubeflow-uats/tree/track/1.10), but we still wanted to ensure that the CI is green here. After the backport is completed, we would merge the changes

.github/workflows/deploy-to-microk8s.yaml
mvlassis: Update for consistency in https://github.com/canonical/charmed-kubeflow-solutions/pull/96/commits/1f9b5c890bb50725270c25f944e78289aba1cea9

.github/dependencies.yaml
deusebio: Yes, I believe you guys have a point. Yes, let's keep it in. I believe I was thinking about this given in TF repo we ONLY have track branches, but having this configurable in a single place it is actually quite convenient.

Also in the MinIO PR (#92 ) I could have used it, and it would have made things way more convenient

deusebio (approved):
Looks good! Thank you for following up on the various comments!

.github/dependencies.yaml
MattiaSarti: Perfect, then!

mvlassis:
After discussing with @deusebio and @MattiaSarti, we are also adding a matrix strategy for running the UATs in AKS and EKS, both with `baseline` and `privileged` security policies. Note that the locations for the the CNI bin and the CNI configuration we are using the default values specified in [the upstream CNI chart](https://github.com/istio/istio/blob/b6ed6e1a7d92be86dbf51fc3d42ab02f1b0e2941/manifests/charts/istio-cni/values.yaml#L22-L23)

kimwnasptd (approved):

PR #102, opened 2025-12-12
[PRA-7] feature: add Kubeflow - Spark integration Terraform bundle

theoctober19th:
This PR adds the feature for Kubeflow <> Spark integration.

Along with the Terraform modules for the feature, this PR also aims to run the UATs in the CI that validate the functionality of the solution.

All changes in this PR were merged to this feature branch after reviews, so this branch contains the reviewed changes. The reason this branch exists here is to test the latest and greatest features worked on so far on Kubeflow - Spark integration epic, before merging it to the base branch.


deusebio:
I have been testing this successfully on PS7, therefore behind a proxy. 

To get it working I followed the guidance provided in the link above, with small tuning to account for the proxy:
* Adding proxy setting to the controller model: 
```
$ juju model-config http-proxy='http://egress.ps7.internal:3128' https-proxy='http://egress.ps7.internal:3128' snap-http-proxy='http://egress.ps7.internal:3128' snap-https-proxy='http://egress.ps7.internal:3128'
```
* Set proxy settings when juju models are created in the main `kubeflow` solution. I raised a PR to add this config properly #104
* Used the following `tfvars.json`:
```
{
    "metacontroller_operator_revision": 551,
    "resource_dispatcher_revision": 410,
    "dex_static_username": "admin",
    "dex_static_password": "admin",
    "http_proxy": "http://egress.ps7.internal:3128",
    "https_proxy": "http://egress.ps7.internal:3128",
    "no_proxy": "10.0.0.0/8"
}
```
* When installing python packages (using `pip`) in the notebooks, the envs proxies should be set (unless they are already set)
```
import os

os.environ["HTTP_PROXY"]="http://egress.ps7.internal:3128"
os.environ["HTTPS_PROXY"]="http://egress.ps7.internal:3128"
```
* Be aware that when proxies are set as above, `spark8t` may not work if the calls to the K8s api are directed to the proxy as well. The `NO_PROXY` setting above should prevent this, but I have figured out that the `httpx` package that `lightkube` depends on does not handle IP ranges (with CIDRs) (see https://github.com/encode/httpx/discussions/2737) , so the no_proxy should be pinning the exact IP of the kubernetes API. This was critical for KFP. Proxy needs to be set (since the KFP stages they generally add a `python3 -m ensurepip` command at startup), but needs to be disabled when `spark8t` is used. To do so, one needs to (I have raised a [PR with these changes](https://github.com/theoctober19th/kubeflow-spark-poc/pull/1/changes)
):
    * add a wrapper `add proxy`, like in [KF UATs](https://github.com/canonical/charmed-kubeflow-uats/blob/main/tests/notebooks/cpu/kfp_v2/kfp-v2-integration.ipynb?short_path=8e4efc3#L94)
    * decorate the KFP component, using the `add_proxy` wrapper
    * reset the NO_PROXY env variable before using `spark8t` (this could possibly be embedded in the `spark8t` utility)

deusebio:
Oh, just an additional comment of an issue I found when closing down the sessions. The executors connected to the Jupyter Spark notebooks keeps lingering around, unless one does stop the context in the notebook using `sc.stop()`. 

If one forcefully close the `Notebook server` or even shut down the Notebook Kernel, the executors are still around e.g.

```
$ kubectl get pod -n admin
NAME                                                          READY   STATUS      RESTARTS   AGE
...
pysparkshell-fec56a9b360f0251-exec-1                          1/2     NotReady    0          4m2s
pysparkshell-fec56a9b360f0251-exec-2                          1/2     NotReady    0          4m
spark-notebook-0                                              2/2     Running     0          4m37s
...
```

Further inspection (using `kubectl describe pod ...`) shows that the issue is due to the istio-side car container to still be running, while the `spark` container is correctly terminated

theoctober19th:
> I have been testing this successfully on PS7, therefore behind a proxy.

Thank you so much @deusebio  for the detailed test in the PS7. This test and the insight you have left in this comment is very much valuable for us. 


> To get it working I followed the guidance provided in the link above, with small tuning to account for the proxy:
> 
>     * Adding proxy setting to the controller model:
> 
> 
> ```
> $ juju model-config http-proxy='http://egress.ps7.internal:3128' https-proxy='http://egress.ps7.internal:3128' snap-http-proxy='http://egress.ps7.internal:3128' snap-https-proxy='http://egress.ps7.internal:3128'
> ```
> 
>     * Set proxy settings when juju models are created in the main `kubeflow` solution. I raised a PR to add this config properly [[MISC] Adding proxy configuration #104](https://github.com/canonical/charmed-kubeflow-solutions/pull/104)


I've cherry-picked the commits in that PR to this PR itself, and now the proxy settings set in the variables should be passed to the Juju model thus created.


>     * Used the following `tfvars.json`:
> 
> 
> ```
> {
>     "metacontroller_operator_revision": 551,
>     "resource_dispatcher_revision": 410,
>     "dex_static_username": "admin",
>     "dex_static_password": "admin",
>     "http_proxy": "http://egress.ps7.internal:3128",
>     "https_proxy": "http://egress.ps7.internal:3128",
>     "no_proxy": "10.0.0.0/8"
> }
> ```

I've also added a new variables file `tfvars-ps7.json` in this PR itself, such that users that want to replicate this in PS7 environment can use this variable file instead.

>     * When installing python packages (using `pip`) in the notebooks, the envs proxies should be set (unless they are already set)
> 
> 
> ```
> import os
> 
> os.environ["HTTP_PROXY"]="http://egress.ps7.internal:3128"
> os.environ["HTTPS_PROXY"]="http://egress.ps7.internal:3128"
> ```
> 
>     * Be aware that when proxies are set as above, `spark8t` may not work if the calls to the K8s api are directed to the proxy as well. The `NO_PROXY` setting above should prevent this, but I have figured out that the `httpx` package that `lightkube` depends on does not handle IP ranges (with CIDRs) (see [NO_PROXY is not working for IP range/mask encode/httpx#2737](https://github.com/encode/httpx/discussions/2737)) , so the no_proxy should be pinning the exact IP of the kubernetes API. This was critical for KFP. Proxy needs to be set (since the KFP stages they generally add a `python3 -m ensurepip` command at startup), but needs to be disabled when `spark8t` is used. To do so, one needs to (I have raised a [PR with these changes](https://github.com/theoctober19th/kubeflow-spark-poc/pull/1/changes)
>       ):
>       
>       * add a wrapper `add proxy`, like in [KF UATs](https://github.com/canonical/charmed-kubeflow-uats/blob/main/tests/notebooks/cpu/kfp_v2/kfp-v2-integration.ipynb?short_path=8e4efc3#L94)
>       * decorate the KFP component, using the `add_proxy` wrapper
>       * reset the NO_PROXY env variable before using `spark8t` (this could possibly be embedded in the `spark8t` utility)

For this, however, I have a slightly different opinion. AFAIK, the `ensurepip` command should not need the proxy (correct if I'm wrong?) because it would be using the scripts that is bundled with the python runtime itself? If we need the proxy behavior only when the user wants to `pip install something`, maybe we enable proxy settings only for those parts of the code that installs some package? This will eliminate the overhead of enabling proxy holistically and then again disabling it for spark8t calls later in the code. 

However, we need to document this such that the user when doing `pip install abcd` will always do so after setting the proxy variables. This is a little cumbersome, but we can write a wrapper for this, and ask the user to use the wrapper whenever necessary.

Either way, I think both cases are easily handled via the instructions we write for the POC (since I don't think we are going to patch spark8t right away to reset `NO_PROXY` anyway), I'd go forward with editing the instructions to make this more clear. And perhaps we can keep both versions of the notebook file in the POC, such that users can look on the example on how to enable / disable proxy when required.

theoctober19th:
I've added the [corresponding notebook files](https://github.com/theoctober19th/kubeflow-spark-poc/blob/main/test_spark_pipeline_proxy.ipynb) for both cases of PS7 behind proxy and environments where there are no proxy. Similarly, I have added separate tfvars example for the case of PS7. Finally, I've also updated the instructions [here](https://github.com/theoctober19th/kubeflow-spark-poc/blob/main/README.md) to include what to do in case of PS7 environment. This should gather everything we've done in this POC so far.

theoctober19th:
Hi @deusebio, I think it's now time for us to merge this feature branch `poc/kubeflow-spark` that contains everything we have done so far for the Kubeflow - Spark integration feature to the base branch `track/1.10`. 

The changes here have passed one round of review already, but nevertheless, let me know if there's anything that you'd rather have it changed in this pull request.

modules/kubeflow-spark/examples/tfvars.json
```diff
@@ -0,0 +1,6 @@
+{
+    "metacontroller_operator_revision": 551,
+    "resource_dispatcher_revision": 410,
```
deusebio: **question** is this required if we would use edge channels? have these revisions been published to any channel? If so, I would just provide the channel in the TF module and not this pinning

deusebio (approved):
LGTM. Just a minor, but I don't think I need to re-rewiew. Hence I'm approving. As far as I'm concerned, Feel free to merge once the comment/discussion is addressed.

modules/kubeflow-spark/examples/tfvars.json
theoctober19th: Yes, the reason we have these pinned here is that these revisions are published to `latest/edge`, whereas the terraform module have their channel set to `risk=stable` by default. Changing the `risk` variable changes the risk for all charms. I had to choose between hardcoding the channel name in the modules, vs have them overridden here, and chos the latter.

theoctober19th: One thing to be noted is that should we change this in the Terraform module itself, we would need to modify the Kubeflow module (which is what I don't want to do). This is because `metacontroller_operator` is deployed by the `kubeflow` module.

deusebio: Yes, right. I agree with you that we don't want to change the Kubeflow module, but we can maybe set the `risk=edge` in the TF spark module. We say that this is experimental, so I don't think this would be strange. 

Maybe just raise another PR on top of this change to see whether this works, and the tests pass. If they don't pass because of the various changes that we are doing to the charms, we can definitely pin these revisions for now.

theoctober19th: Setting `risk=edge` only in TF Spark module would only cover the case for `resource-dispatcher`. We also need to override the risk for `metacontroller-operator`, which is deployed via `kubeflow` module. 

Perhaps we can move these variables from here, to the `kubeflow-spark` module itself, hardcoding the value when it is passed to `kubeflow` module -- however I don't think that's a good idea. To be honest, having these overrides here would perhaps also be a cue for someone using this to immediately understand this is an experimental feature, and also remind us to publish the charm to stable revisions every time we see these variables.

theoctober19th:
**TODO** Revert [f1fdf35](https://github.com/canonical/charmed-kubeflow-solutions/pull/102/commits/f1fdf350525dde198e997b797b56415fe895423e) before merging, once the tests pass and CI is green.

theoctober19th (commented):
Waiting for the CI to turn green to proceed.

modules/kubeflow-spark/examples/tfvars.json
deusebio: I see that you have raised the PR: https://github.com/canonical/charmed-kubeflow-solutions/pull/120, and unfortunately it fails

So I believe it is fine to merge with the revisions here, but we should then aim at getting the other PR green once that the KF bundle is overall more stable. When doing so, we should remove the revisions here and update the documentation, see [here](https://github.com/canonical/charmed-kubeflow-documentation/pull/34/changes#r2741299078)

deusebio: Resolving the thread here

modules/kubeflow/main.tf
```diff
@@ -1,6 +1,12 @@
 resource "juju_model" "kubeflow" {
   count = var.create_model ? 1 : 0
   name  = local.model
+
+  config = {
+    juju-http-proxy  = var.http_proxy
+    juju-https-proxy = var.https_proxy
+    juju-no-proxy    = var.no_proxy
+  }
```
MattiaSarti, line 9: question: how did we test this works? The CI does not cover the case where a proxy is needed

modules/kubeflow-spark/tests/test_deployment.py
```diff
@@ -0,0 +1,89 @@
...
+    @pytest.mark.dependency(depends=["TestCharm::test_apply_terraform_solution"])
+    async def test_assert_deployment(self, ops_test: OpsTest, lightkube_client):
+        """
+        Wait for the applications to become active and idle and verify its public URL access.
+        """
+
+        apps = list(ops_test.model.applications.keys())
+        
+        # Remove opentelemetry-collector-k8s-kubeflow from the apps list because it remains
+        # `blocked` until it's related to one of the COS charms
```
MattiaSarti, line 44: praise: helpful comment

.github/workflows/deploy-to-k8s.yaml
MattiaSarti, line 1: question: don't we want to run Spark's UATs in AWS and EKS too?

.github/dependencies.yaml
```diff
@@ -1,4 +1,4 @@
 CKF_BRANCH: "track/1.10"  
 K8S_VERSION: "1.32"
 JUJU_VERSION: "3.6"
-UATS_BRANCH: "track/1.10"
+UATS_BRANCH: "main"
```
MattiaSarti: todo: let's restore this before merging (I know we needed it for testing), but... 

issue: ...if we do so, won't the CI on `track/1.10` fail, then? I would suggest merging once we have the UAT changes backported, because we can't have the solutions on `track/1.10` non-aligned with the respective UATs

MattiaSarti (commented):
Excellent work!

Food for thought for @deusebio: soon or later, we may want to invest some effort to refactor the submodules so that the whole logic for deploying, which is 99% identical across submodules, is defined in a unified place (otherwise, every time we need to remember to make changes in multiple places)

.github/dependencies.yaml
theoctober19th: Oh yes, that's a valid point. Either we need to backport the UATs to `track/1.10` (but I don't think that would be a good idea because we are not releasing the Kubeflow Spark module on track 1.10).

Or perhaps we can patch our CI such that Spark tests are run on `main` while other tests use the usual `track/1.10` for test.

Let's hear what @deusebio has to say on this.

deusebio: Very good point! I believe we should 

> patch our CI such that Spark tests are run on main while other tests use the usual track/1.10 for test.

The right way to do this would have been to have a "main" for the TF solution as well. But we don't, so let's just hard code the UAT branch for the Spark tests here

deusebio:
> Food for thought for @deusebio: soon or later, we may want to invest some effort to refactor the submodules so that the whole logic for deploying, which is 99% identical across submodules, is defined in a unified place (otherwise, every time we need to remember to make changes in multiple places)

Yes I believe this is already thought of happening in the TF refactoring epic. We also have a sort of [spec about this](https://docs.google.com/document/d/1BBuslhuG38v4TfepQJUuy0Ijk1NKu7_PklWKxVZ8kgM/edit?tab=t.0)

modules/kubeflow/main.tf
deusebio: Yes, we did test on PS7 and reported the findings [here](https://github.com/canonical/charmed-kubeflow-solutions/pull/102#issuecomment-3674333238)

.github/workflows/deploy-to-k8s.yaml
deusebio: My 2cents: as long as they are experimental, it is fine to keep them only on certain substrates. When they become GA and stable, yes, we should

MattiaSarti: Sure, it makes sense.

modules/kubeflow/main.tf
MattiaSarti: Got it, thanks for the reference!

.github/dependencies.yaml
theoctober19th: I've fixed this in the PR #120, merging it shortly after the CI is green.

MattiaSarti (approved):
Perfect!

note: even if the CI was red before restoring the branch of the UATs, [the pull request merged into this branch](https://github.com/canonical/charmed-kubeflow-solutions/pull/120) with the changes of the failing CI run had its CI pass all tests before being merged, so that is proof the CI was green even before reverting the UATs' branch

theoctober19th:
Thanks for the reviews! The CI is fully green now after the commit [5127efd](https://github.com/canonical/charmed-kubeflow-solutions/pull/102/commits/5127efdd68e55ab127de1814d9d0ee0b53332968), and thus proceeding to revert the `--bundle ""` changes done earlier in the PR via commit [8e5dadc](https://github.com/canonical/charmed-kubeflow-solutions/pull/102/commits/8e5dadc858bc28ec623ddddbbaa402da0aa5aade).

theoctober19th:
After reverting the change `--bundle ""` to disable bundle correctness test temporarily, the three Spark tests are now failing (expectedly). However since we verified they pass by disabling bundle correctness test in [5127efd](https://github.com/canonical/charmed-kubeflow-solutions/pull/102/commits/5127efdd68e55ab127de1814d9d0ee0b53332968), I'm now proceeding to merge this PR in it's current state.

cc: @MattiaSarti @deusebio

PR #104, opened 2025-12-18
[MISC] Adding proxy configuration

deusebio:


theoctober19th:
Thanks for the fix, @deusebio! Since the PR is not mergeable due to failures at the moment, I am going to copy these changes to the other branch via cherry-pick.

PR #110, opened 2026-01-14
[PRA-88] Add UAT tests to Spark <> Kubeflow Integration feature branch

theoctober19th:
This PR adds the Kubeflow <> Spark UAT tests to the **POC feature branch**, utilizing the UATs we have written in https://github.com/canonical/charmed-kubeflow-uats/pull/230.

We run the UAT tests for both microk8s and Canonical K8s here.


.github/workflows/deploy-to-k8s.yaml
```diff
@@ -31,11 +31,17 @@ jobs:
...
             uats:
               env: uats-remote
-              extra-args: --filter "not feast" --include-kubeflow-trainer-tests
+              extra-args: --filter "not feast and not spark" --include-kubeflow-trainer-tests
           - module: kubeflow-feast
             uats:
               env: feast-remote
               extra-args: ""
+          - module: kubeflow-spark
+            juju-agent-version: 3.6.9
```
deusebio: **niptick, out-of-scope** There has *just* been a new Juju release that should fix this, but let's not address this here. Let's just have a ticket on the Spark side to fix this. As part of the ticket we can also try to bump the version here and see whether things still work. 

In this context, just attach this comment to the ticket for removing the 3.6.9 pinning in Spark. That should be enough to resolve this comment.

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -32,11 +33,17 @@ jobs:
...
             uats:
               env: uats-remote
-              extra-args: --filter "not feast" --include-kubeflow-trainer-tests
+              extra-args: --filter "not feast and not spark" --include-kubeflow-trainer-tests
           - module: kubeflow-feast
             uats:
               env: feast-remote
               extra-args: ""
+          - module: kubeflow-spark
+            juju-agent-version: 3.6.9
```
deusebio: **niptick, out-of-scope** as above.

.github/concierge.yaml
```diff
@@ -1,5 +1,6 @@
 juju:
   channel: 3.6/stable
+  agent-version: 3.6.9
```
deusebio: **niptick, out-of-scope** as above. Just link this comment to the ticket

deusebio (approved):
LGTM! Just place let's make sure that removing the pinning is part of the current scope for removing the 3.6.9 pinning

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -8,6 +8,7 @@ on:
       uats_branch:
         description: 'Branch to run the UATs from e.g. main or track/1.10. By default, this is defined by the dependencies.yaml file.'
         required: false
+        default: 'poc/kubeflow-spark'
```
deusebio: **todo** Update the channel. Keep the comment open until the PR on UATs is merged, and then just address it ahead of merging.

Keep the comment open such that we make sure that this does not fall through the cracks

.github/dependencies.yaml
```diff
@@ -1,4 +1,4 @@
 CKF_BRANCH: "track/1.10"  
 K8S_VERSION: "1.32"
 JUJU_VERSION: "3.6"
-UATS_BRANCH: "track/1.10"
+UATS_BRANCH: "poc/kubeflow-spark"
```
deusebio: **todo** update channel

.github/workflows/deploy-to-k8s.yaml
theoctober19th: Created a new issue #114 from this comment.

dariofaccin (approved):

.github/workflows/deploy-to-microk8s.yaml
theoctober19th: Updated in [745762e](https://github.com/canonical/charmed-kubeflow-solutions/pull/110/commits/745762e070eb0143d5d6fdf9e695fa4eb9044d70).

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -31,11 +31,17 @@ jobs:
...
             uats:
               env: uats-remote
-              extra-args: --filter "not feast" --include-kubeflow-trainer-tests
+              extra-args: --filter "not feast and not spark" --include-kubeflow-trainer-tests
           - module: kubeflow-feast
             uats:
               env: feast-remote
               extra-args: ""
+          - module: kubeflow-spark
+            juju-agent-version: 3.6.9 # TODO: remove pin once Spark can work on recent Juju agent versions
```
mvlassis: nit: We don't actually use this here, since the `agent-version` value in `concierge.yaml` is used. We can delete this line here and keep the comment in `.github/concierge.yaml`.
```suggestion
```

modules/kubeflow-spark/main.tf
```diff
@@ -1,75 +1,77 @@
...
+  oidc_gatekeeper_ca_bundle             = var.oidc_gatekeeper_ca_bundle
+  pvcviewer_operator_revision           = var.pvcviewer_operator_revision
+  tensorboard_controller_revision       = var.tensorboard_controller_revision
+  tensorboards_web_app_revision         = var.tensorboards_web_app_revision
+  training_operator_revision            = var.training_operator_revision
 }
 
 module "resource_dispatcher" {
-  source     = "git::https://github.com/canonical/resource-dispatcher//terraform?ref=main"
+  source     = "git::https://github.com/canonical/resource-dispatcher//terraform?ref=893c73d48f49023f0cf3aa13927a609167d53bf7"
```
mvlassis, line 74: question: Are we pinning this hash for a specific reason? It seems that the necessary changes we need (roles/role_bindings) have been merged to main in https://github.com/canonical/resource-dispatcher/pull/135

mvlassis (commented):
Nice stuff! Left a few minor comments to disambiguate

.github/workflows/deploy-to-k8s.yaml
theoctober19th: You're right. I've removed this line in the next commit.

modules/kubeflow-spark/main.tf
theoctober19th: `tflint` complains if the pinning is done with a default branch like `main` -- that's the reason the pin here is a SHA with the latest commit on the main branch.

This doesn't happen with the other modules, because they pin `track/1.10`, something different from the default branch like `main`.

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -31,11 +31,16 @@ jobs:
...
+              extra-args: --filter "not feast and not spark" --include-kubeflow-trainer-tests
           - module: kubeflow-feast
             uats:
               env: feast-remote
               extra-args: ""
+          - module: kubeflow-spark
+            tf-vars-file: examples/tfvars.json
+            uats:
+              env: spark-remote
+              extra-args: "--test-image ghcr.io/canonical/charmed-spark-jupyterlab:3.5-22.04_edge"
```
misohu, line 43: Why is this edge ? when will we have stable image ?

modules/kubeflow-spark/main.tf
misohu: But the plan is to release new version and then use the track branch right ?

misohu (approved):
Small non blocking comments

modules/kubeflow-spark/main.tf
theoctober19th: Yes, eventually the charm will be released on a track and the track branch reference be used instead.

deusebio: I don't mind too much right now since this is going to be refactored, but using branches is not a good practice and it makes a TF module not reproducible. Using floating point references (i.e. branches) has been breaking deployments in the past, and we should avoid it. It is also part of the new [TF standard](https://docs.google.com/document/d/1k1psLCfcf0Nr5KeVtWGFi9wmzhcg5AP9GVZTsRyVQSQ/edit?tab=t.0)

.github/workflows/deploy-to-k8s.yaml
deusebio: we can also just reference to a particular hash, such that we will then move the hash to stable once this is consolidated and supported

deusebio: Actually, now that I'm thinking about it, being these the tests actually using edge makes sense such that we spot regression straight away. I believe it is fine to use edge

mvlassis (approved):

Issue #108, opened 2026-01-12
Update application versions for release 1.11


dariofaccin:
### Context

In order to release Kubeflow `1.11`, some charms need to be updated. Assuming charms and application terraform modules are updated, the solutions repository needs to be updated to deploy the correct version of such applications.

### What needs to get done

1. KServe version update to `0.16`
2. KFP version update to `2.15`
3. Katib version update to `0.19`
4. Istio version update to `1.28`

### Definition of Done

1. KServe application is updated to `0.16`
2. KFP application is updated to `2.15`
3. Katib application is updated to `0.19`
4. Istio application is updated to `1.28`

PR #119, opened 2026-01-27
chore: update references for 1.11/edge kubeflow

misohu:
Part of: https://github.com/canonical/charmed-kubeflow-solutions/issues/108


.github/workflows/deploy-to-k8s.yaml
```diff
@@ -14,7 +14,7 @@ on:
       risk:
         description: 'The risk level to be tested'
         required: false
-        default: 'stable'
+        default: 'edge'
```
MattiaSarti: suggestion: shall we revert this change? It is unnecessary because inputs are used for manual dispatches, while in our CI what counts is the content of `dependencies.yaml`, which we already updated to `edge` risk and which we already read from environment variables (from what I had experienced, you will see ${{ inputs.risk }} not defined in CI runs)

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -14,7 +14,7 @@ on:
       risk:
         description: 'The risk level to be tested'
         required: false
-        default: 'stable'
+        default: 'edge'
```
MattiaSarti: suggestion: shall we revert this change? It is unnecessary because inputs are used for manual dispatches, while in our CI what counts is the content of `dependencies.yaml`, which we already updated to `edge` risk and which we already read from environment variables (from what I had experienced, you will see ${{ inputs.risk }} not defined in CI runs)

modules/kubeflow/variables.tf
```diff
@@ -1,7 +1,7 @@
 variable "risk" {
   type        = string
   description = "Value for the risk to be used"
-  default     = "stable"
+  default     = "edge"
```
MattiaSarti, line 4: question: why is this necessary? If it's only for our CI runs, it should be already overridden by the deployment command, being read from the environment variable

MattiaSarti (changes requested):
Nice! Just a couple of details:

mvlassis:
Note that we are pining the version of `pydantic` to `<2.11` in the `e2e-wine` test due to https://github.com/canonical/charmed-kubeflow-uats/issues/218

modules/kubeflow/variables.tf
misohu: Which command you mean ?

.github/workflows/deploy-to-microk8s.yaml
misohu: Yep you right i will do it after we publish all new edge charms :)

modules/kubeflow/variables.tf
MattiaSarti: [this](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.10/.github/workflows/deploy-to-microk8s.yaml#L161) :arrow_right: [this](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.10/modules/kubeflow-mlflow/tests/conftest.py#L50) :arrow_right: [this](https://github.com/canonical/charmed-kubeflow-solutions/blob/track/1.10/modules/kubeflow-mlflow/tests/test_deployment.py#L27-L31)

.github/workflows/deploy-to-k8s.yaml
misohu: ok i would keep this candidate for now as we will send this change for further testing if they set it to stable it will break (as there are no stable versions for the target charms)

.github/workflows/deploy-to-microk8s.yaml
misohu: ok i would keep this candidate for now as we will send this change for further testing if they set it to stable it will break (as there are no stable versions for the target charms)

modules/kubeflow/variables.tf
misohu: ok i would keep this candidate for now as we will send this change for further testing if they set it to stable it will break (as there are no stable versions for the target charms)

.github/workflows/deploy-to-k8s.yaml
MattiaSarti: No problem.

.github/workflows/deploy-to-microk8s.yaml
MattiaSarti: No problem.

modules/kubeflow/variables.tf
MattiaSarti: No problem.

deusebio:
I noticed that there is some issues on deployments given to the fact that for resource dispatcher and feast charms were missing the `risk=candidate` revision. Apparently, the TF Juju provider is not able to resolve the revisions when the risk is not published. 

I have now published:
* the revision `547` of resource dispatcher to `2.0/candidate`
* the revision `136` of feast-ui to `0.49/candidate` and `0.49/beta`
* the revision `165` of feast-integrator to `0.49/candidate` and `0.49/beta`

so hopefully this will fix the CI

modules/kubeflow/variables.tf
MattiaSarti: suggestion (non-blocking): this can still be undone, though. It is overridden by what set in `dependencies.yaml`.

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -14,7 +14,7 @@ on:
       risk:
         description: 'The risk level to be tested'
         required: false
-        default: 'stable'
+        default: 'candidate'
```
MattiaSarti, line 17: suggestion (non-blocking): this can still be undone, though. It is overridden by what set in `dependencies.yaml`.

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -14,7 +14,7 @@ on:
       risk:
         description: 'The risk level to be tested'
         required: false
-        default: 'stable'
+        default: 'candidate'
```
MattiaSarti, line 17: suggestion (non-blocking): this can still be undone, though. It is overridden by what set in `dependencies.yaml`.

.github/workflows/deploy-to-k8s.yaml
MattiaSarti: This can still be undone, though. It is overridden by what set in `dependencies.yaml`.

.github/workflows/deploy-to-microk8s.yaml
MattiaSarti: This can still be undone, though. It is overridden by what set in `dependencies.yaml`.

MattiaSarti (approved):

deusebio:
I believe it is fine to merge as-is. We can raise a dedicated PR to swap back to "Stable" and make sure the env variable is set correctly, as @MattiaSarti pointed out

deusebio (approved):
Looks good!

.github/workflows/deploy-to-k8s.yaml
deusebio: part of https://warthogs.atlassian.net/browse/KF-8398

.github/workflows/deploy-to-microk8s.yaml
deusebio: part of https://warthogs.atlassian.net/browse/KF-8398

modules/kubeflow/variables.tf
deusebio: part of https://warthogs.atlassian.net/browse/KF-8398

.github/workflows/deploy-to-k8s.yaml
deusebio: part of https://warthogs.atlassian.net/browse/KF-8398

.github/workflows/deploy-to-microk8s.yaml
deusebio: part of https://warthogs.atlassian.net/browse/KF-8398

PR #120, opened 2026-01-30
[test] Test the integration with Kubeflow from edge

theoctober19th:
Test the integration tests with risk=edge, removing the revision pinning.

### DIFF
--- a/.github/workflows/deploy-to-k8s.yaml
+++ b/.github/workflows/deploy-to-k8s.yaml
@@ -38,7 +38,8 @@ jobs:
               extra-args: --bundle ""
           - module: kubeflow-spark
             tf-vars-file: examples/tfvars.json
-            risk: edge
+            risk: edge # TODO: Run Spark tests with Kubeflow from `stable` risk once all changes for Spark integration are released to stable
+            uats_branch: main # TODO: Run Spark tests against ${{ inputs.uats_branch }} once UATs backported to track/1.x
             uats:
               env: spark-remote
               extra-args: --test-image ghcr.io/canonical/charmed-spark-jupyterlab:3.5-22.04_edge@sha256:72a6e89985e35e0920fb40c063b3287425760ebf823b129a87143d5ec0e99af7 --bundle ""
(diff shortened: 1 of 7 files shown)


deusebio:
When this becomes stable, we should definitely remember to also update the documenation here: https://github.com/canonical/charmed-kubeflow-documentation/pull/34/changes#r2741299078

modules/kubeflow-feast/applications.tf
```diff
@@ -16,7 +16,7 @@ module "feast_ui" {
 
 module "feast_offline_store" {
   # tflint-ignore: terraform_module_pinned_source
-  source          = "git::https://github.com/canonical/postgresql-k8s-operator//terraform?ref=main"
+  source          = "git::https://github.com/canonical/postgresql-k8s-operator//terraform?ref=ac1f27eae4d9ccee2557ec84a175fbb5722e45be" # TODO: Unpin after juju TF provider upgraded to >1.0
```
theoctober19th, line 19: Had to pin these because Postgresql Terraform modules upgraded to use juju provider 1.0 just an hour ago, but our modules expect juju provider <1.0 -- thus resulting in test failure.

This will be unpinned once we make our modules compatible with juju TF provider > 1.0

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -38,6 +38,8 @@ jobs:
               extra-args: --bundle ""
           - module: kubeflow-spark
             tf-vars-file: examples/tfvars.json
+            risk: edge # TODO: Run Spark tests with Kubeflow from `stable` risk once all changes for Spark integration are released to stable
+            uats_branch: main # TODO: Run Spark tests against ${{ inputs.uats_branch }} once UATs backported to track/1.x
```
deusebio, line 42: **praise** very nice!

modules/kubeflow-feast/applications.tf
deusebio: Nice!

deusebio (approved):
Looks good! Thank you for also fixing the tf-provider v<1.0.0 pinning

MattiaSarti (approved):

PR #130, opened 2026-02-26
fix: adjust the dry-run behavior and file parsing

misohu:
- Cleaned up [charms_promotions.py](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and removed stale commented code.
- Added explicit --dry-run / --apply flags (dry-run stays default).
- Dry-run now prints real [charmcraft release ...](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) commands instead of Python list output.
- Fixed text parsing so full juju status output works (including model/controller section).
- Added warning+skip behavior for charms where charmcraft status fails due to missing permissions, instead of aborting the whole run.
- Added a small usage note in [README.md](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,37 @@
+# Scripts
+
+## `charms_promotions.py`
+
+Promotes charm revisions from a Juju status export to a higher risk channel (`beta` → `candidate` → `stable`).
+
+### What it reads
+
+- `--format text`: a `juju status` text output (full output is fine, not just the apps table)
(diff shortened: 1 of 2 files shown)


deusebio (approved):
All changes look good! Also I like to enforce either `dry-run` or `apply`. I believe it is more explicit

PR #159, opened 2026-04-16
feat: add ambient component

### LINKED ISSUE #147: Create `Istio` components
### Context

Assuming [exploration](https://github.com/canonical/charmed-kubeflow-solutions/issues/143) on components is completed, the `Istio` components need to be created.
The idea is to have two different components: one including `istio-k8s`, `istio-ingress-k8s` and `istio-beacon-k8s`; the other `istio-pilot` and `istio-gateway`.

### What needs to get done

1. Create `Istio` components.

### Definition of Done

1. `Istio` components are created.

dariofaccin:
This PR adds the `istio` ambient component.
The component includes three applications, `istio-k8s`, `istio-ingress-k8s` and `istio-beacon-k8s`.
The intra-component relation between them (`istio-ingress-config`) is defined in `integrations.tf`.
The provided relations (`gateway`, `service-mesh`, `istio-ingress-route`, `istio-ingress-route-unauthenticated`) are outputs and defined in `outputs.tf`.
The required relations (`forward-auth`) are defined in `outputs.tf`.

This PR closes #147


terraform-refactoring/components/ambient/variables.tf
```diff
@@ -0,0 +1,55 @@
...
+    revision    = optional(number)
+    units       = optional(number, 1)
+    trust       = optional(bool, true)
+    constraints = optional(string)
+    resources   = optional(map(string), {})
+  })
+  default = {}
+}
+
+variable "istio_k8s_platform" {
```
deusebio: **question** why are we lifting the config option from the istio-k8s charm into the root level? I believe this pattern may make sense when we need to make sure consistency of certain parameters across the component, e.g. one variable that controls multiple config options across the various components. However here it seems that this value is then just fed to the config option of only one charm. So I don't think it is really worth breaking the standard/guidance

terraform-refactoring/components/ambient/applications.tf
```diff
@@ -0,0 +1,55 @@
...
+    channel  = var.istio_k8s.channel
+    revision = var.istio_k8s.revision
+  }
+
+  model_uuid  = var.model_uuid
+  name        = "istio-k8s"
+  units       = var.istio_k8s.units
+  trust       = var.istio_k8s.trust
+  constraints = var.istio_k8s.constraints
+  config = var.istio_k8s_platform != "" ? {
```
deusebio: **question** what if the user needs to specify some of the other configs in the `istio-k8s` charm? Do we want to protect to set the other config options? I'm also concern if in the future there will be more config. I believe we should just expose the config as we do for other charms, and possibly surfacing high-level variable only at product level

terraform-refactoring/components/ambient/README.md
```diff
@@ -0,0 +1,80 @@
...
+  }
+}
+```
+
+## Inputs
+
+| Name | Type | Default | Description |
+|---|---|---|---|
+| `model_uuid` | `string` | required | UUID of the Juju model |
+| `risk` | `string` | `"edge"` | Channel risk: `stable`, `candidate`, `beta`, or `edge` |
```
deusebio: **todo** remove, as this is not there any more

terraform-refactoring/components/ambient/README.md
```diff
@@ -0,0 +1,80 @@
...
+- Kubernetes cluster with CNI support for ambient mesh
+
+## Usage
+
+```hcl
+module "ambient" {
+  source = "../../components/ambient"
+
+  model_uuid = juju_model.kubeflow.uuid
+  risk       = "stable"
```
deusebio: **todo** remove

deusebio (changes requested):
I have some concerns about lifting the platfrom variable from the charm variable to a dedicated standalone variable. I'm thinking that this pattern may make sense ONLY when there is a high-level variable that controls multiple config that must be aligned. A bit like what you were doing with `risk` (although as we have possibly discussed, that is a good variable rather for the product)

terraform-refactoring/components/ambient/variables.tf
dariofaccin: I think this was part of the first iteration and never updated.

terraform-refactoring/components/ambient/applications.tf
dariofaccin: See comment above regarding the `istio_k8s_platform` exposure. Will fix in next commit.

deusebio (approved):
Looks good! The only comment here is that we should probably eventually donate this component module to the service-mesh team. It feels something they should eventually own. Let's move ahead with this, but also use Madrid to sync with them to figure out whether they would like to onboard this into their repository, and for us to reference that

PR #158, opened 2026-04-16
feat: add istio sidecar component

dariofaccin:
This PR adds the `istio` sidecar component.
The component includes two applications, `istio-pilot` and `istio-gateway`.
The intra-component relation between them is defined in `integrations.tf`.
The provided relations (`gateway`, `ingress`, `ingress-auth`) are outputs and defined in `outputs.tf`.

This PR closes #147

Issue #143, opened 2026-04-09
Explore components splitting for Terraform modules refactoring


dariofaccin:
### Context

The scope of this task is to understand which charms can be bundled in the same Component entity, which integrations are intra-component and which are inter-components and need to be exposed.

### What needs to get done

1. Understand which charms can be bundled in the same `Component` entity
2. Determine which integrations are intra-component
3. Determine which integrations are inter-components and need to be exposed

### Definition of Done

1. Component entities are outlined
2. Intra-component relations are outlined
3. Inter-component relations are outlined

dariofaccin:
The following diagrams outline how Charmed Kubeflow could be split into `Components`: each component has its own life-cycle.
The diagrams only show the _intra-component integrations_: these integrations will not be exposed as outputs.

# Core
```mermaid
flowchart LR
  admission_webhook["admission-webhook"]
  argo_controller["argo-controller\n(canonical/argo-operators)"]
  dex_auth["dex-auth"]
  envoy["envoy"]
  kubeflow_dashboard["kubeflow-dashboard"]
  kubeflow_profiles["kubeflow-profiles"]
  kubeflow_roles["kubeflow-roles"]
  kubeflow_volumes["kubeflow-volumes"]
  metacontroller["metacontroller"]
  mlmd["mlmd"]
  minio["minio"]
  oidc_gatekeeper["oidc-gatekeeper"]
  pvcviewer_operator["pvcviewer"]

  dex_auth -->|dex_oidc_config| oidc_gatekeeper
  oidc_gatekeeper -->|oidc_client| dex_auth
  kubeflow_profiles -->|kubeflow_profiles| kubeflow_dashboard
  kubeflow_dashboard -->|dashboard_links| kubeflow_volumes
  mlmd -->|grpc| envoy
```


# Katib
```mermaid
flowchart LR
  katib_controller["katib-controller"]
  katib_db_manager["katib-db-manager"]
  katib_ui["katib-ui"]

  katib_db_manager -->|k8s_service_info| katib_controller
```

# KFP
```mermaid
flowchart LR
  kfp_api["kfp-api"]
  kfp_metadata_writer["kfp-metadata-writer"]
  kfp_persistence["kfp-persistence"]
  kfp_profile_controller["kfp-profile-controller"]
  kfp_schedwf["kfp-schedwf"]
  kfp_ui["kfp-ui"]
  kfp_viewer["kfp-viewer"]
[...]

PR #162, opened 2026-04-17
feat: add core component

### LINKED ISSUE #144: Create `Core` component
### Context

Assuming [exploration](https://github.com/canonical/charmed-kubeflow-solutions/issues/143) on components is completed, the `Core` component needs to be created.

### What needs to get done

1. Create `Core` component.

### Definition of Done

1. `Core` component is created.

dariofaccin:
This PR adds the `core` component.
This PR also refactors the testing environment so that a single `tox` configuration handles all refactored components.
The environment includes utilities to format all modules according to coding style and a script to generate/update READMEs of all modules via [terraform-docs](https://terraform-docs.io/).

**Note**: this PR depends on #158 and #159.

This PR closes: #144


terraform-refactoring/components/core/applications.tf
```diff
@@ -0,0 +1,104 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+# MinIO object storage application
+resource "juju_application" "minio" {
```
deusebio: **thought** I'm thinking that we may want to leave `core` "stateless", and rather have this externally to the core module. Minio could rather be standalone `charm-module` that eventually we could swap to the `s3-integrator`, that's also why I feel this may need to be externalized

But let's not over-engineer this just yet. So it is fine to keep it here for the time being, but let's have this in the back of our minds.

deusebio (commented):
Looks good! Just a small non-blocking comment for minio, that we may also end up to address in a future refactoring. 

The only more major comment I would have is that IF the core module can work standalone (e.g. going into a active/idle state), we could probably also start working on the product modules that wire up the various components. For now only the core ones are there, but this way we can also have a simple CI to test that when you deploy everthing is active idle, and check consistency as we progress

We can of course also address this in a dedicated ticket if you prefer, but I would start pushing things together sooner rather than later

terraform-refactoring/components/core/applications.tf
dariofaccin: I moved `minio` as a separate charm and integrated it back at the product level.

deusebio (approved):
Looks good! Non-blocking, but I would just possibly add a `juju status` in the CI, such that we see easily what's being deployed. I went to the CI and looked at the ugly TF logs, which are quite unreadable. If you are still working on this PR to address other comments, please factor that in. Otherwise, it is also fine to address this in the next PR or some work-in-progress that you have

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -120,40 +117,23 @@ jobs:
...
-          git clone https://github.com/canonical/charmed-kubeflow-uats.git ~/charmed-kubeflow-uats
-          cd ~/charmed-kubeflow-uats
-          git checkout ${{ env.UATS_BRANCH }}
-          tox -e ${{ matrix.bundle.uats.env }} -- ${{ matrix.bundle.uats.extra-args }}
-
-      # On failure, capture debugging resources
-      - name: Select model
-        run: juju switch kubeflow
-        if: failure() || cancelled()
+          tox -c ./terraform-refactoring/products/kubeflow -e test_deployment -- -vv -s $RISK_ARG $PLATFORM_ARG $SERVICE_MESH_TYPE_ARG
```
MattiaSarti, line 128: nitpick: if `kubeflow-ambient` is not run as an actual product subfolder, in the sense that both sidecar mode and ambient mode are part of `products/kubeflow` and are simply parametrized with `--service-mesh-type`, then shall we rename the GitHub matrix from `product: kubeflow` and `product: kubeflow-ambient` to something else, like `service-mesh-type: sidecar` and `service-mesh-type: ambient`?

scripts/tox.ini
MattiaSarti, line 1: praise: tidy, lovely parametrization

scripts/generate-docs.sh
```diff
@@ -0,0 +1,29 @@
...
+TERRAFORM_DIR=$(pwd)/../terraform-refactoring
+
+echo "Starting terraform-docs update from ${TERRAFORM_DIR}"
+
+find "${TERRAFORM_DIR}" -maxdepth 2 -mindepth 2 -type d | while read -r MODULE_DIR; do
+  echo "Processing ${MODULE_DIR}"
+  if [ -d "${MODULE_DIR}" ]; then
+    cd "${MODULE_DIR}"
+
+    if [ -f "README.md" ]; then
```
MattiaSarti: nitpick: if it's a valid module, don't we want to generate docs even when there is no `README.md` yet?

terraform-refactoring/charms/minio/outputs.tf
MattiaSarti, line 1: todo: we'll eventually have to update `provides` and `requires`, but we can also do that on the final pull request and not in this one

terraform-refactoring/components/core/applications.tf
MattiaSarti, line 1: issue: [here](https://github.com/canonical/charmed-kubeflow-solutions/issues/143) we designed the core component to include `pvc-viewer`, which made sense since it's the UI of `kubeflow-volumes` (as far as I know, at least), but we didn't end up including it in this pull request

terraform-refactoring/products/kubeflow/tests/conftest.py
```diff
@@ -0,0 +1,57 @@
...
+        default="microk8s",
+        type=str,
+        help="Platform for istio-k8s (e.g., microk8s, or empty string for Canonical K8s)",
+    )
+    parser.addoption(
+        "--service-mesh-type",
+        nargs="?",
+        default="istio",
+        type=str,
+        help="Service mesh type (istio or ambient)",
```
MattiaSarti: nitpick: to remove ambiguity, shall we make it `sidecar` and `ambient`, instead of `istio` and `ambient`?

terraform-refactoring/products/kubeflow/main.tf
MattiaSarti, line 1: question: why do we use URL-based `source` for some modules and relative path-based `source` for others?

terraform-refactoring/components/core/outputs.tf
MattiaSarti, line 1: todo: we'll eventually have to update `provides` and `requires`, but we can also do that on the final pull request and not in this one

terraform-refactoring/components/core/integrations.tf
MattiaSarti, line 1: issue: I see a couple integrations missing:
- Volumes ↔️ Dashboard
- the ones of PVCViewer (see comment above)

MattiaSarti (changes requested):

scripts/generate-docs.sh
dariofaccin: I will try to address this by checking if any `.tf` files are defined in the current folder.

terraform-refactoring/components/core/applications.tf
dariofaccin: Nice catch, I will add `pvc-viewer` charm too.

terraform-refactoring/products/kubeflow/main.tf
dariofaccin: I am using local `source` for modules that are not merged yet and part of this PR (only `core` in this PR). I could also make them point to this actual branch, but it doesn't change much.
On subsequent PRs I will replace the local source with the URL pointing to `feat/terraform-refactoring` branch for the merged components.

terraform-refactoring/components/core/outputs.tf
dariofaccin: I would address COS-related endpoints in a separate PR once all components are ready, otherwise the back and forth will be a lot.

terraform-refactoring/charms/minio/outputs.tf
dariofaccin: I would address COS-related endpoints in a separate PR once all components are ready, otherwise the back and forth will be a lot.

.github/workflows/deploy-to-k8s.yaml
dariofaccin: Currently we only support Kubeflow product.
As part of the epic, also MLFlow and Feast will be introduced back so I think it makes sense to maintain this "useless" (for now) parametrization.

terraform-refactoring/products/kubeflow/main.tf
MattiaSarti: Then it makes perfect sense! Sure

.github/workflows/deploy-to-k8s.yaml
MattiaSarti: Okay, no problem, but when new products are introduced, then can we remember to have the `service-mesh-type` as a subfield for each product, instead of having doubled products for sidecar and ambient modes?

MattiaSarti (approved):

PR #163, opened 2026-04-17
feat: add kubeflow product

### LINKED ISSUE #151: Create `Kubeflow` product
### Context

Assuming all components are created, the `Kubeflow` product needs to be created.

### What needs to get done

1. Create `Kubeflow` product.

### Definition of Done

1. `Kubeflow` product is created.

dariofaccin:
This PR completes the minimal working version of refactored Kubeflow product.
This PR only introduces a variable named `release` which is used to pin the tracks of the charm applications; valid values are `1.11` and `latest`.

**Note**: matrix testing of `mlflow-ambient` is disabled due to [this issue](https://github.com/canonical/mlflow-operator/issues/410).

Note: `terraform` snap revision is pinned to `v1.14.9` due to a bug in Terraform `v1.15.0`.

This PR closes #151.


.github/workflows/deploy-to-k8s.yaml
```diff
@@ -114,7 +118,7 @@ jobs:
 
       - name: Setup terraform
         run: |
-          sudo snap install terraform  --channel latest/stable --classic
+          sudo snap install terraform --revision=859 --classic
```
MattiaSarti, line 121: question: why this pinning? In case, do we need to do the same on MicroK8s?

terraform-refactoring/products/kubeflow/variables.tf
```diff
@@ -3,6 +3,17 @@
 
 # Juju Settings
 
+variable "release" {
+  type        = string
+  description = "Kubeflow release to deploy. Use 'latest' for latest tracks or '1.11' for pinned 1.11 tracks."
+  default     = "latest"
+
+  validation {
+    condition     = contains(["1.11", "latest"], var.release)
```
MattiaSarti, line 12: suggestion (non-blocking): how about using a boolean, so as to make the `var.release == "1.11"` checks more maintainable without repeating the string value to check against? Or are we planning on expanding this to more than two options in the future?

MattiaSarti (approved):

MattiaSarti (approved):

.github/workflows/deploy-to-k8s.yaml
dariofaccin: The pinning is temporary due to a bug in the `v1.15.0` version of terraform. The upstream fix is already available in `v1.15.1`, but I am not expecting a snap upgrade this week, so I pinned to the latest `v1.14.9` revision available.

terraform-refactoring/products/kubeflow/variables.tf
dariofaccin: The idea is to support two stable releases, currently `1.10` and `1.11`, plus the possibility to deploy from `latest`.
I could actually address the addition of `1.10` tracks in this PR actually (in that case, a new validation needs to be added to prevent deploying `1.10` release with ambient mode).

.github/workflows/deploy-to-k8s.yaml
MattiaSarti: Got it

terraform-refactoring/products/kubeflow/variables.tf
MattiaSarti: Got it - no need!

PR #167, opened 2026-04-23
feat: add Katib component

### LINKED ISSUE #146: Create `Katib` component
### Context

Assuming [exploration](https://github.com/canonical/charmed-kubeflow-solutions/issues/143) on components is completed, the `Katib` component needs to be created.

### What needs to get done

1. Create `Katib` component.

### Definition of Done

1. `Katib` component is created.

dariofaccin:
This PR closes #146.


terraform-refactoring/products/kubeflow/tox.ini
MattiaSarti, line 1: suggestion (non-blocking): we could adapt the same `tox.ini` configurations as in our charms after the recent dependency management system refactoring (see [a random one](https://github.com/canonical/kserve-operators/blob/main/charms/kserve-controller/tox.ini)), with these little differences: 

- we don't have to define `poetry` as an external and install it beforehand, we can simply define `tox` as the only external dependency and have `poetry` as a dependency that is automatically installed by `tox` - we currently do this in our charms:
    ```
    allowlist_externals =
    	tox
    deps =
    	poetry>=2.1.3
     ```

- we could always skip installation and install only what under the target environment (but it doesn't really matter if it's in `commands` or in `commands_pre`), even in the integration test environment:
    ```
    commands_pre =
	    poetry install --only integration
    skip_install = true
    ```

terraform-refactoring/components/katib/outputs.tf
MattiaSarti, line 1: todo: usual reminder to eventually update `provides` and requires` on the target branch of the pull request

MattiaSarti (approved):

PR #168, opened 2026-04-23
chore: switch to `latest/edge`

### LINKED ISSUE #164: Upgrade KServe to `0.17.0`
### Context

KServe in undergoing an upgrade to `0.17.0`.

### What needs to get done

Upgrade the charm for KServe to `0.17.0`.

### Definition of Done

The charm for KServe is upgraded to `0.17.0`.

MattiaSarti:
This pull request switches all MLOps charms to `latest/edge` channels while also parametrizing the bundle-correctness check for Istio's mode.

_note: this automatically upgrades KServe to `0.17.0` and validates it with the UATs (KServe `0.17.0` is only available on `latest` track and on `edge` risk at the time of writing)_

resolves https://github.com/canonical/charmed-kubeflow-solutions/issues/164


.github/workflows/deploy-to-k8s.yaml
```diff
@@ -132,7 +132,7 @@ jobs:
           git clone https://github.com/canonical/charmed-kubeflow-uats.git ~/charmed-kubeflow-uats
           cd ~/charmed-kubeflow-uats
           git checkout ${{ env.UATS_BRANCH }}
-          tox -e ${{ matrix.bundle.uats.env }} -- ${{ matrix.bundle.uats.extra-args }}
+          tox -e ${{ matrix.bundle.uats.env }} -- ${{ matrix.bundle.uats.extra-args }} --bundle ""
```
deusebio: **todo** why are we disabling this? Now we should have everything in the same repo, so just updating the reference [here](https://github.com/canonical/charmed-kubeflow-uats/blob/main/assets/versions.yaml) should fix this

deusebio (changes requested):
Looks good! Only one small change to NOT disable the bundle check

.github/workflows/deploy-to-k8s.yaml
MattiaSarti: Right! There is one problem, though: [the versions/tracks of main in the UATs](https://github.com/canonical/charmed-kubeflow-uats/blob/main/assets/versions.yaml) are all `latest` and do not correspond to the versions/tracks on main in the solutions, which this branch is created from, that instead report the same tracks as in CKF 1.11, so I would need to either update all versions/tracks on main in the UATs to be no longer `latest` or to update them on this pull request (and consequently on main of the Terraform solutions) to make them actually point to `latest`. Which option do we prefer to go for? @deusebio

PS: this issue was never spotted because this is the first pull request to the newly created main branch

modules/kubeflow-ambient/applications.tf
```diff
@@ -160,10 +160,10 @@ module "kfp_viz" {
 }
 
 module "kserve_controller" {
-  source     = "git::https://github.com/canonical/kserve-operators//charms/kserve-controller/terraform?ref=main"
+  source     = "git::https://github.com/canonical/kserve-operators//charms/kserve-controller/terraform?ref=track/0.17"
```
deusebio: **question** why are we not using `main` here? I would have expected the opposite :D

deusebio (commented):
Looks very good! Only one small question

modules/kubeflow-ambient/applications.tf
MattiaSarti: Because... it's a typo 🥲 Nice catch, thanks!

NohaIhab (approved):

modules/kubeflow-ambient/applications.tf
deusebio: :joy:

deusebio (approved):
Looks good!

MattiaSarti:
note: merging without waiting for all tests to run again because all tests passed before the final, merge commit

PR #175, opened 2026-04-27
feat: add Training component

### LINKED ISSUE #150: Create `Training` component
### Context

Assuming [exploration](https://github.com/canonical/charmed-kubeflow-solutions/issues/143) on components is completed, the `Training` component needs to be created.

### What needs to get done

1. Create `Training` component.

### Definition of Done

1. `Training` component is created.

dariofaccin:
This PR adds `training-operator` and `kubeflow-trainer` as `Notebooks` component.
This component exposes an input variable named `enable_training_v2` to choose whether to deploy Kubeflow Trainer V2 along with Training Operator V1.
This PR closes #150.


terraform-refactoring/components/training/variables.tf
```diff
@@ -0,0 +1,98 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+variable "model_uuid" {
+  description = "UUID of the Juju model where Training is deployed"
+  type        = string
+  nullable    = false
+}
+
+variable "enable_v2" {
```
deusebio, line 16: **thought** I'm starting to think that maybe we could have both `enable_v1` and `enable_v2`. Eventually, we could transition from deploying by default v1 to deploying by default v2, so I would prepare for that

terraform-refactoring/products/kubeflow/main.tf
```diff
@@ -343,7 +343,7 @@ module "tensorboard" {
   count      = var.enable_tensorboard ? 1 : 0
   depends_on = [module.core, module.istio, module.ambient]
 
-  source = "../../components/tensorboard"
+  source = "git::https://github.com/canonical/charmed-kubeflow-solutions//terraform-refactoring/components/tensorboard?ref=feat/terraform-refactor"
```
deusebio, line 346: **question** why we are not using the relative import? hasn't the PR already landed?

deusebio (approved):
Looks good! Just some minor, non-blocking comments

terraform-refactoring/products/kubeflow/main.tf
dariofaccin: There is actually no change between using relative import from this branch or git source from `feat/terraform-refactor` branch. I am using relative imports only for components not merged into the feature branch or that are changed by the PR itself.

terraform-refactoring/components/training/variables.tf
dariofaccin: Added `enable_v1` flag, set to `true` by default.

deusebio: Thanks!!!

terraform-refactoring/products/kubeflow/main.tf
deusebio: Yes yes, it makes sense. It was just a question

deusebio (approved):
Looks good!

PR #176, opened 2026-04-27
feat: add KServe and KNative component

### LINKED ISSUE #149: Create `KServe` component
### Context

Assuming [exploration](https://github.com/canonical/charmed-kubeflow-solutions/issues/143) on components is completed, the `KServe` component needs to be created.

### What needs to get done

1. Create `KServe` component.

### Definition of Done

1. `KServe` component is created.

dariofaccin:
This PR adds KServe and KNative component.
KNative operators are only deployed when `gateway-info` variable is populated (e.g., istio-sidecar mode).
This PR closes #149.


deusebio (approved):
Looks good! Now that I see this PR, I realized that there may be some synergies of this work together with the work that @MattiaSarti has been doing to upgrade the [autoscaling-serving-bundle](https://github.com/canonical/autoscaling-model-serving/pull/17). 

Actually, I'm just thinking if under `products`, we could also have one standalone module that only deploys a similar serving solution as the one in the other repo/PR, while still leveraging on the same KNative/KServe TF component. That's of course fine to address this in a separate PR, where we try to reproduce the changes that Mattia has done. But I feel this is definitely worth considering, in the aim at eventually deprecating the `autoscaling-serving-bundle` and only maintaining the TF modules here

deusebio (commented):
Removed the approval, since the CI does not seem to be happy yet. Could you check if the failure are consistent?

MattiaSarti:
> Looks good! Now that I see this PR, I realized that there may be some synergies of this work together with the work that @MattiaSarti has been doing to upgrade the https://github.com/canonical/autoscaling-model-serving/pull/17.

> Actually, I'm just thinking if under products, we could also have one standalone module that only deploys a similar serving solution as the one in the other repo/PR, while still leveraging on the same KNative/KServe TF component. That's of course fine to address this in a separate PR, where we try to reproduce the changes that Mattia has done. But I feel this is definitely worth considering, in the aim at eventually deprecating the autoscaling-serving-bundle and only maintaining the TF modules here

Sure, @deusebio! It makes sense. So, as you said, we'll introduce this in a future pull request and for now I'll proceed with my work on https://github.com/canonical/autoscaling-model-serving/pull/17, since the timeline for the two tasks is different. But when the time comes, I will be glad to help with that.

deusebio:
Yes yes, that's totally fine to not block either work, since they are anyway orthogonal things as they stand right now

dariofaccin:
@deusebio the CI should be fixed now: I was missing an integration in ambient mode and to properly set `deployment-mode` configuration.

terraform-refactoring/products/kubeflow/main.tf
```diff
@@ -419,7 +425,9 @@ module "kserve" {
   kserve_controller = {
     channel  = local.kserve_channel
     revision = var.kserve_controller_revision
-    config   = var.kserve_controller_config
+    config = merge({
+      "deployment-mode" = var.service_mesh_type == "sidecar" ? "knative" : "standard"
```
deusebio, line 429: **though** just checking: basically here we provide a default that if not specified otherwise, the deployment mode is "knative" for sidecar and "standard" for ambient. I'm just wondering: would it be possible to deploy "standard" with "sidecar"? to do so, could I feed the `kserve_controller_config` with `deployment-mode: standard`, that would override the default setting?

deusebio (approved):
Looks good! I only have a small question, but the response should not be blocking. If the default behaviour cannot be modified, maybe it will be better to have two variables (one to control the istio mode, and one to control the deployment-mode of kserve. We could eventually also provide a validation that "knative" with "ambient" is not allowed. 

 But this can also be approached as fine tuning of the product module to be done later.

terraform-refactoring/products/kubeflow/main.tf
dariofaccin: I created a `tfvars` file with the following content:
```
service_mesh_type = "sidecar"

kserve_controller_config = {
  "deployment-mode" = "standard"
}
```
The output of `terraform apply` is the following:
```
# module.kserve[0].juju_application.kserve_controller will be created
  + resource "juju_application" "kserve_controller" {
      + config      = {
          + "deployment-mode" = "standard"
        }
      + constraints = "arch=amd64"
      + id          = (known after apply)
      + machines    = (known after apply)
      + model_type  = (known after apply)
      + model_uuid  = (known after apply)
      + name        = "kserve-controller"
      + resources   = {}
      + storage     = (known after apply)
      + trust       = true
      + units       = 1

      + charm {
          + base     = (known after apply)
          + channel  = "latest/edge"
          + name     = "kserve-controller"
          + revision = (known after apply)
        }
    }
```
So the answer is yes, the default is overridden.

dariofaccin:
@deusebio I added a couple of validations, so that `deployment-mode` cannot be set to `knative` when `service_mesh_type` is `ambient`, and viceversa (`standard` with `sidecar`).
The outputs are like this:
```
╷
│ Error: Invalid value for variable
│ 
│   on variables.tfvars line 5:
│    5: kserve_controller_config = {
│    6:   "deployment-mode" = "standard"
│    7: }
│     ├────────────────
│     │ var.kserve_controller_config["deployment-mode"] is "standard"
│     │ var.service_mesh_type is "sidecar"
│ 
│ deployment-mode cannot be set to 'standard' when service_mesh_type is 'sidecar'.
│ 
│ This was checked by the validation rule at variables.tf:480,3-13.
╵
```
```
╷
│ Error: Invalid value for variable
│ 
│   on variables.tfvars line 5:
│    5: kserve_controller_config = {
│    6:   "deployment-mode" = "knative"
│    7: }
│     ├────────────────
│     │ var.kserve_controller_config["deployment-mode"] is "knative"
│     │ var.service_mesh_type is "ambient"
│ 
│ deployment-mode cannot be set to 'knative' when service_mesh_type is 'ambient'.
│ 
│ This was checked by the validation rule at variables.tf:475,3-13.
╵
```

deusebio (commented):
> viceversa (standard with sidecar).

The validation looks good, but I don't think we should prevent sidecar to NOT work with standard. Provided the bug that @MattiaSarti has found is addressed, probably this combination will be supported as well!

Otherwise, the code changes here look very good and sane! Please also just fix the linting such that I can do the last round of review!

deusebio (approved):
Looks good!

PR #177, opened 2026-04-28
feat: add MLFlow component

dariofaccin:
This PR adds MLFlow component and resource dispatcher charm.
A variable `enable_mlflow` dictates whether to deploy `resource-dispatcher` and MLFlow.
The PR adds a new local, `deploy_kserve` which is set to true when at least one between `enable_mlflow` or `enable_kserve` is true, since KServe is required by MLFlow.
The PR also refactors other locals so that `mysql` and `minio` are only deployed when required.

This PR closes #171.

PR #180, opened 2026-04-30
feat: add Feast component

### LINKED ISSUE #172: Create `Feast` component
### Context

Assuming [exploration](https://github.com/canonical/charmed-kubeflow-solutions/issues/143) on components is completed, the `Feast` component needs to be created.

### What needs to get done

1. Create `Feast` component.

### Definition of Done

1. `Feast` component is created.

dariofaccin:
This PR adds Feast components and Postgresql charm.
A variable `enable_feast` dictates whether to deploy `resource-dispatcher`, Feast and postgresql.
The PR adds a new local, `deploy_kserve` which is set to true when at least one between `enable_mlflow` or `enable_kserve` is true, since KServe is required by MLFlow.
The PR also refactors other locals so that `mysql` and `minio` are only deployed when required.

Note: this PR is created on top of #177 to include `resource-dispatcher`.

This PR closes #172.


terraform-refactoring/charms/resource-dispatcher/outputs.tf
MattiaSarti, line 1: todo: usual reminder to update `provides` and `requires` outputs

terraform-refactoring/components/feast/outputs.tf
MattiaSarti, line 1: todo: usual reminder to update `provides` and `requires` outputs

terraform-refactoring/components/feast/integrations.tf
MattiaSarti, line 1: issue: there are a couple integrations missing:
- a similar one for feast-ui to provide an ingress as we have it here but for ambient mode, instead of sidecar mode
- ambient-mode service-mesh relation to include feast-ui
- ambient-mode service-mesh relation to include resource-dispatcher

These are missing even in the current Terraform solutions at the time of writing, but are required to run Feast in ambient mode

MattiaSarti (changes requested):

terraform-refactoring/components/feast/integrations.tf
dariofaccin: The ambient integration for `resource-dispatcher` is handled at product level, since it is a charm module (and not a component module).
Added missing integrations (`istio-ingress-route` and `service-mesh`) for `feast-ui`.

deusebio: Nice catch! Yes, `resource-dispatcher` is a tricky one. I'm thinking about this, and I believe that eventually it may be bundled to the kubeflow component. Its role seems rather tight to the existence of profiles and metacontroller, and providing ability to *implement* user-namespace integrations with external components (MLFlow, Databases, Spark, Feast, etc). Exposing low-level K8s resources integrations (such as secrets, pod-defaults, etc) seems to fine-grained and low-level for a product API. On the other hand, I'm thinking that a better integration boundary could be at the `kubeflow-data-integrator` level, such that what the `kubeflow` product can actually expose are the high-level endpoints of that charm (see [here](https://charmhub.io/data-kubeflow-integrator/integrations)). 

This will require a bit of discussion and design. So definitely out of scope for this PR. But please folks, have a thought about this. BTW @MattiaSarti : maybe mediating the relation between MLFlow and resource-dispacther using `kubeflow-data-integrator` could also be the way to enable multi-tenancy, and therefore the `kubeflow-data-integrator` and `resource-dispatcher` may be central components to enable multi-tenancy anyway, so worth adding to the kubeflow core component

deusebio (approved):
Looks good! Just a long comment, which is more for us to think about, rather than something to be addressed in the context of this PR

terraform-refactoring/components/feast/integrations.tf
dariofaccin: @deusebio let me know if you want `resource-dispatcher` to be always deployed as part of _Core_ component. The drawback of this approach is that the Feast and MLFlow components then will be **fully dependent** on Core component.

MattiaSarti: Thank you both for the brainstorming, I see the pros and cons for both alternatives.

MattiaSarti (approved):

PR #182, opened 2026-05-04
feat: add COS integration

dariofaccin:
This PR adds COS integration to Kubeflow product by creating a new Deployment, named _Kubeflow-COS_.
A new component, named `observability`, is created, which deploys `opentelemetry-collector-k8s` and performs the proper integrations with Kubeflow components.
At deployment level, two module are deployed: COS, provided by observability-team, and Kubeflow. The offers of COS are used as inputs for the Kubeflow product (these will be consumed by the observability component inside).

**NOTE**: the deployment is currently not deployed as part of the CI, since it requires fine tuning of storage configuration to avoid flakiness.

### MinIO Overview
minio

### Kubeflow dashboard - Node.js process overview
kf_nodejs

### Kubeflow Dashboard - Latency, traffic, errors & saturation
kf_dashboard

### Jupyter Notebook Controller
notebooks

### Example logs - Dex
loki_logs

### Example metrics - Envoy
envoy_metrics


terraform-refactoring/components/kserve/outputs.tf
MattiaSarti, line 1: praise: nice parametrizations

terraform-refactoring/components/observability/integrations.tf
MattiaSarti, line 1: issue: a couple of integrations might be missing:
- knative-serving -> `otel_collector`
- knative-eventing -> `otel_collector`

MattiaSarti (changes requested):

MattiaSarti (approved):

scripts/tox.ini
```diff
@@ -39,19 +41,22 @@ set_env =
...
     minio: MODULE=minio
     mlflow: MODULE=mlflow
     notebooks: MODULE=notebooks
+    observability: MODULE=observability
     resource-dispatcher: MODULE=resource-dispatcher
     tensorboard: MODULE=tensorboard
+    training: MODULE=training
 
-[testenv:lint-{auth,core,istio-ambient,istio-sidecar,feast,katib,kubeflow,minio,mlflow,resource-dispatcher,tensorboard}]
+[testenv:lint-{auth,core,istio-ambient,istio-sidecar,feast,katib,kubeflow,kubeflow-cos,minio,mlflow,notebooks,observability,minio,mlflow,resource-dispatcher,tensorboard}]
```
mvlassis: nitpick: We add `minio` and `mlflow` here twice, right?
```suggestion
[testenv:lint-{auth,core,istio-ambient,istio-sidecar,feast,katib,kubeflow,kubeflow-cos,minio,mlflow,notebooks,observability,resource-dispatcher,tensorboard}]
```

scripts/tox.ini
```diff
@@ -39,19 +41,22 @@ set_env =
...
 description = Check Terraform code against coding style standards
 commands =
     terraform -chdir={[vars]tf_modules_path}{env:TYPE}/{env:MODULE} fmt -recursive -check -diff
     tflint --recursive --chdir={[vars]tf_modules_path}{env:TYPE}/{env:MODULE}
 
-[testenv:validate-{auth,core,istio-ambient,istio-sidecar,feast,katib,kfp,kubeflow,minio,mlflow,notebooks,resource-dispatcher,tensorboard,training}]
+[testenv:validate-{auth,core,istio-ambient,istio-sidecar,feast,katib,kfp,kubeflow,kubeflow-cos,minio,mlflow,notebooks,observability,minio,mlflow,notebooks,resource-dispatcher,tensorboard,training}]
```
mvlassis: nitpick: Same as above, we add `minio`,`mlflow` twice here too
```suggestion
[testenv:validate-{auth,core,istio-ambient,istio-sidecar,feast,katib,kfp,kubeflow,kubeflow-cos,minio,mlflow,notebooks,observability,notebooks,resource-dispatcher,tensorboard,training}]
```

scripts/tox.ini
```diff
@@ -61,7 +66,7 @@ commands =
 description = Apply coding style standards to code
 commands =
     terraform -chdir={[vars]tf_modules_path} fmt -recursive
-    tflint --recursive --fix --chdir={[vars]tf_modules_path}{env:TYPE}/{env:MODULE}
+    tflint --recursive --fix --chdir={[vars]tf_modules_path}
```
mvlassis, line 79: praise: That is an excellent catch, nice one!

mvlassis (commented):
Left a small nitpick

terraform-refactoring/components/kserve/outputs.tf
```diff
@@ -13,24 +13,70 @@ output "components" {
...
+output "provides" {
+  description = "Map of endpoints provided by this component to other components (outbound relations)"
+  value = merge(
+    {
+      kserve_controller_metrics_endpoint = {
+        name     = juju_application.kserve_controller.name
+        endpoint = "metrics-endpoint"
+      }
+    },
+    length(juju_application.knative_operator) > 0 ? {
```
deusebio: **suggestion** maybe we could use a local variable for this use case that also centralize whether knative_operator should be deployed or not, e.g. 

```
locals {
  mode = "serverless" or "standard" # this depending on the input variable, like the `var.gateway` or similar
}
```

and then 

```
resource "juju_application" "knative_serving" {
  count = locals.mode == "serverless" ? 1 : 0
  ...
```

and similarly here

```
output "provides" {
  description = "Map of endpoints provided by this component to other components (outbound relations)"
  value = merge(
    locals.mode == "serverless" ? {
```

This would make the code a bit more readable, I believe

deusebio (commented):
Indeed, very good and solid work @dariofaccin ! The structure here is very much consistent with how the integration with observability is implemented also for Spark. Now that the things are pretty much in place, it could be good to get Alex to have a look at the code base

I only have a minor niptick. Also, I would probably rather move the `deployment` folder into the tests, instead of the tf main modules. Speaking of which, are we trying to deploy the "Deployment" anywhere?

terraform-refactoring/components/kserve/outputs.tf
dariofaccin: Updated with a similar approach.

dariofaccin:
@deusebio I refactored the modules and tests so that:
* `kubeflow-cos` does not live with other modules (e.g., as a Deployment) but is under tests
* CI now deploys Kubeflow + COS (and MLFlow / Feast according to matrix)

deusebio:
It seems that the CI is timing out. I believe I have seen this in the past, when the storages requested by all charms were going beyond the capacity. When this was happening, then one of the application (in this case mlmd) keeps on hanging because it never gets the storage, and the CI times out. Now that also COS is in the picture, some of those charms may also request storage which goes beyond the allocated capacity available to Canonical K8s. 

One possible resolution could be to just configure the storage for the charms to smaller values, e.g. `1GB` instead of `10GB`. 

But also, I believe it is fine to have just one tests testing the integration between KF and COS, where we could also use larger runners if needed. I would in general not always deploy KF + COS, as this may make the CI a bit more flaky, but up to you.

Batalex (commented):
I don't have much to say about COS itself (except that I'm jealous you can use otelcol, seems like the loki integration is not working well for us). 

I do have a general remark about OCI resources though, which is that we should expose a variable to pin a specific OCI revision any time we provide a way to pin a charm revision. Otherwise, you risk deploying an older, tested charm revision while pulling the very latest workload OCI resource, which is not a 100% reproducible setup

dariofaccin:
@Batalex thanks for the input. Yes, I realized I missed the OCI resources exposure when the PR was almost done. I would issue another PR, since some clean-up and refinement is expected anyway, to address those.
@deusebio if you are ok with this I merge.

deusebio (approved):
Yes sure, that fine. Let's address this in a separate PR

Issue #173, opened 2026-04-27
Upgrade Terraform bundle to use version `2.16.0` of kfp


mvlassis:
### Context

.

### What needs to get done

.

### Definition of Done

.

PR #184, opened 2026-05-06
feat: Update kfp charms to `2.16` in release candidate branch

mvlassis:
## Summary
This PR updates the `kubeflow` module to include track `2.16` of the 8 kfp charms. Note that since we haven't released to `2.16/stable` yet, we're using the `2.16/edge` channel.

Part of #173

### DIFF
--- a/.github/dependencies.yaml
+++ b/.github/dependencies.yaml
@@ -1,5 +1,5 @@
 CKF_BRANCH: "track/1.11-rc"
 K8S_VERSION: "1.32"
 JUJU_VERSION: "3.6"
-UATS_BRANCH: "track/1.11-rc"
+UATS_BRANCH: "kf-8655-rc-branch"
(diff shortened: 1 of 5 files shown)


MattiaSarti (changes requested):
todo: instead of skipping the bundle-correctness check, let's update the respective version file(s) on the UAT's side

deusebio:
> todo: instead of skipping the bundle-correctness check, let's update the respective version file(s) on the UAT's side

Good point. Maybe we can just have an track/1.11-rc on UATs as well. Or for an RC branch (given it is a feature branch), it is also fine to merge a `--bundle ""` (such that we are not blocked), provided this will be removed before the final release

MattiaSarti (approved):

MattiaSarti (approved):

PR #189, opened 2026-05-25
[KF-8759] feat: Kubeflow <> Spark integration

deusebio:
The PR proposes:
* a skeleton to add external integration coming from kubeflow-integrator and the bindings to external product, e.g. postgresql, opensearch, etc
* The skeleton is then applied to allow Spark integration with the new format
* CI is updated to do UATs for Spark

The PR also fixes some smaller issues, most notably some incorrect integration between knative and otel-collector


terraform-refactoring/components/data-kubeflow-integrator/main.tf
deusebio, line 1: **comment** initially it was a charm module (hence the position), but I believe it now looks more like a component, since the codebase here is less "standard" and provides some extra high-level functionalities/inputs/outputs (e.g. parametric creation of external integrations) differently from charm modules.

terraform-refactoring/components/observability/integrations.tf
```diff
@@ -901,55 +901,6 @@ resource "juju_integration" "knative_operator_logging" {
   }
 }
 
-# ---------------------------------------------------------------------------
-# OTel collector integrations (knative apps → opentelemetry-collector-k8s)
-# ---------------------------------------------------------------------------
```
deusebio, line 906: **issue** This has been giving me error when deploying and I'm not sure what is the "otel-collector" endpoint? was it an allucination?

terraform-refactoring/products/kubeflow/variables.tf
```diff
@@ -517,6 +517,39 @@ variable "enable_mlflow" {
   default     = false
 }
 
+# Enable integrations
+
+variable "integrations" {
```
deusebio: **highlight** I'm very much keen to have feedback on the API to allow external integration. The idea is that this should be an high-level input to deploy various `data-kubeflow-integrator` instances

terraform-refactoring/tests/rag/main.tf
```diff
@@ -0,0 +1,98 @@
...
+  enable_tensorboard = false
+  enable_training_v1 = false
+  enable_training_v2 = false
+  enable_mlflow      = true
+  enable_kserve      = true
+  enable_feast       = false
+
+  kserve_controller_config = {
+    deployment-mode="standard"
+    custom_images="{\"serving_runtimes__huggingfaceserver\": \"kserve/huggingfaceserver:v0.17.1\", \"serving_runtimes__huggingfaceserver__multinode\": \"kserve/huggingfaceserver:v0.17.1\"}"
```
deusebio: **issue** our image of "huggingface" is not properly working for GPUs. It seems that we are missing some C libraries. I'll be raising a bug, but meanwhile we need to use upstream.

terraform-refactoring/components/data-kubeflow-integrator/main.tf
deusebio: refactored and moved to components.

terraform-refactoring/components/observability/integrations.tf
deusebio: I believe this has been allucinating. Interenstingly I have probably understood why copilot got it wrong. In the charms there was already an otel-collector endpoints, but it was between knative-controller <> knative-eventing and knative-controller <> knative-serving. 

When refactoring, copilot probably thought it was supposed to be with the opentelemetry collector. 

I have re-added the missing integration for the knative below, see [here](https://github.com/canonical/charmed-kubeflow-solutions/pull/189/changes#diff-8b6f15de7ad966ddf6a4e59dc1b4d718827334fa9692892980a3bec2f82efe61R1013)

terraform-refactoring/tests/rag/main.tf
deusebio: rag module has been removed from this repo, and moved into https://github.com/deusebio/charmed-retrival-augment-generation#

terraform-refactoring/products/kubeflow/main.tf
```diff
@@ -707,6 +726,101 @@ module "feast" {
...
+    secret-key = var.s3_secret_key
+    access-key = var.s3_access_key
+  }
+  info = "This is the access key and secret key for the S3 storage"
+}
+
+module "s3" {
+  depends_on = [juju_model.kubeflow, juju_secret.s3_secret]
+  count      = var.enable_spark ? 1 : 0
+  source     = "git::https://github.com/canonical/spark-k8s-bundle//terraform/charms/s3-integrator?ref=wip-split-components"
```
deusebio: **todo** update once the PR in spark is merged (reference: https://github.com/canonical/spark-k8s-bundle/pull/233)

terraform-refactoring/products/kubeflow/variables.tf
```diff
@@ -14,6 +14,17 @@ variable "release" {
   }
 }
 
+variable "profile" {
```
deusebio, line 17: **note** This is a pattern we have been using for DP components, when tuning is required to deploy in constraint environment, e.g. see [Kafka config options](https://charmhub.io/kafka-k8s/configurations#profile). 

There is also a spec outlining more the rationale: https://docs.google.com/document/d/1bq7KyUAPIIcUWGgifW05p40v7lGXs3PhMznM2ECaMKY/edit?tab=t.0

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -112,6 +120,35 @@ jobs:
...
+          lxc init ubuntu:jammy ceph -c limits.cpu=4 -c limits.memory=2GB -d root,size=5GB
+          
+          lxc config set ceph cloud-init.user-data - < microceph_rgw.yaml
+          lxc start ceph
+          while ! lxc exec ceph -- id -u ubuntu &>/dev/null; do sleep 0.5; done
+          lxc exec ceph -- cloud-init status --wait
+          
+          echo -e "S3_SERVER_URL=http://$(lxc list --format json | yq '.[] | select(.name == "ceph") .state.network.eth0.addresses.[] | select(.family == "inet") .address'):80/\nS3_ACCESS_KEY=foo\nS3_SECRET_KEY=bar" > .env
+          lxc list
+          cat .env
```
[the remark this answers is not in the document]
deusebio: Not needed, this are test credentials from the local microceph deployment. It is good for debugging. Let's leave it in.

terraform-refactoring/products/kubeflow/variables.tf
```diff
@@ -725,3 +777,112 @@ variable "postgresql_k8s_config" {
...
+}
+
+# S3 Integrator variables
+
+variable "s3_secret_key" {
+  description = "S3 secret key for object storage integration"
+  type        = string
+  default     = ""
+  sensitive   = true
+}
```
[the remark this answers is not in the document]
deusebio: Not needed

terraform-refactoring/products/kubeflow/variables.tf
```diff
@@ -725,3 +777,112 @@ variable "postgresql_k8s_config" {
...
+
+variable "s3_access_key" {
+  description = "S3 access key for object storage integration"
+  type        = string
+  default     = ""
+  sensitive   = true
+}
```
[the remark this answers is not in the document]
deusebio: not needed

terraform-refactoring/products/kubeflow/variables.tf
```diff
@@ -725,3 +777,112 @@ variable "postgresql_k8s_config" {
...
+
+variable "s3_endpoint" {
+  description = "S3 endpoint for object storage integration"
+  type        = string
+  default     = ""
+}
+
```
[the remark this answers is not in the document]
deusebio: not needed

terraform-refactoring/products/kubeflow/variables.tf
```diff
@@ -725,3 +777,112 @@ variable "postgresql_k8s_config" {
...
+variable "s3_bucket" {
+  description = "S3 bucket for object storage integration"
+  type        = string
+  default     = ""
+}
+
```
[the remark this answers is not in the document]
deusebio: not needed

terraform-refactoring/tests/conftest.py
```diff
@@ -81,15 +90,31 @@ def enable_feast(request) -> list[str]:
...
+    """Terraform module customization for Spark deployment."""
+    if request.config.getoption("--enable-spark"):        
+        extra_args = [
+            "-var", "enable_spark=true", 
+            "-var", "s3_bucket=spark-demo",
+            "-var", f"s3_secret_key={os.environ.get('S3_SECRET_KEY')}",
+            "-var", f"s3_access_key={os.environ.get('S3_ACCESS_KEY')}",
+            "-var", f"s3_endpoint={os.environ.get('S3_SERVER_URL')}",
+        ]
+        print(f"Extra args for Spark deployment: {extra_args}")
```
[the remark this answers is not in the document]
deusebio: Not needed, this are test credentials from the local microceph deployment. It is good for debugging. Let's leave it in.

terraform-refactoring/tests/conftest.py
```diff
@@ -81,15 +90,31 @@ def enable_feast(request) -> list[str]:
     return []
 
 @pytest.fixture(scope="module")
-def tf_vars(risk, service_mesh_type, enable_mlflow, enable_feast) -> list[str]:
+def enable_spark(request) -> list[str]:
+    """Terraform module customization for Spark deployment."""
+    if request.config.getoption("--enable-spark"):        
+        extra_args = [
+            "-var", "enable_spark=true", 
+            "-var", "s3_bucket=spark-demo",
```
MattiaSarti: suggestion (non-blocking): let's allow the solutions' user to parametrize the S3 bucket name too

terraform-refactoring/tests/conftest.py
```diff
@@ -1,8 +1,12 @@
 import jubilant
 import pytest
 
+from dotenv import load_dotenv
+import os
+
 MODEL_NAME = "kubeflow"
 
+load_dotenv()
```
MattiaSarti, line 9: question: do we expect the solutions' user to have the sensitive environment variables used below in a local `.env` file because they are external to Juju, being them for external S3 storage?

terraform-refactoring/components/observability/integrations.tf
```diff
@@ -901,55 +901,6 @@ resource "juju_integration" "knative_operator_logging" {
...
-
-resource "juju_integration" "knative_operator_otel_collector" {
```
MattiaSarti, line 908: todo: you missed adding back one of the three ones: "knative_operator_otel_collector"

terraform-refactoring/products/kubeflow/main.tf
```diff
@@ -707,6 +726,101 @@ module "feast" {
...
+    resources   = var.spark_integration_hub_image != null ? { integration-hub-image = var.spark_integration_hub_image } : null
+  }
+
+  # Integrations
+
+  object_storage           = merge({ kind = "endpoint" }, module.s3[0].provides.s3_credentials)
+  object_storage_interface = module.s3[0].provides.s3_credentials.endpoint
+}
+
+module "integrations" {
```
MattiaSarti: thought: the name integrations, here for the module and in the other file for the variable, may be misleading to the unfamiliar developer, as they are not direct integrations in Juju-Terraform's sense

terraform-refactoring/products/kubeflow/main.tf
```diff
@@ -453,7 +464,7 @@ module "tensorboard" {
 }
 
 module "resource_dispatcher" {
-  count      = (var.enable_mlflow || var.enable_feast) ? 1 : 0
+  count      = (var.enable_mlflow || var.enable_feast || length(local.integrations) > 0) ? 1 : 0
```
MattiaSarti: praise: nice idea

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -126,19 +163,22 @@ jobs:
           RISK_ARG="--risk ${{ env.RISK }}"
           PLATFORM_ARG=""
           SERVICE_MESH_TYPE_ARG="--service-mesh-type=${{ matrix.service-mesh-type }}"
-          MLFLOW_ARG=""
-          FEAST_ARG=""
+          
+          EXTRA_COMPONENTS_ARG=""
+
           if [[ "${{ matrix.service-mesh-type }}" == "ambient" ]]; then
-            RISK_ARG=""
```
MattiaSarti, line 132: question: why are we removing this? Wasn't it intentional to avoid setting unsupported risks for ambient mode?

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -112,6 +120,35 @@ jobs:
...
+        run: |
+          echo -e "$CLOUD_INIT_FILE" > microceph_rgw.yaml
+          lxc init ubuntu:jammy ceph -c limits.cpu=4 -c limits.memory=2GB -d root,size=5GB
+          
+          lxc config set ceph cloud-init.user-data - < microceph_rgw.yaml
+          lxc start ceph
+          while ! lxc exec ceph -- id -u ubuntu &>/dev/null; do sleep 0.5; done
+          lxc exec ceph -- cloud-init status --wait
+          
+          echo -e "S3_SERVER_URL=http://$(lxc list --format json | yq '.[] | select(.name == "ceph") .state.network.eth0.addresses.[] | select(.family == "inet") .address'):80/\nS3_ACCESS_KEY=foo\nS3_SECRET_KEY=bar" > .env
```
MattiaSarti, line 148: nitpick: we could break it down in 3 lines, one for each environment variable, to then append to the file the second and third times, to make everything more readable

terraform-refactoring/products/kubeflow/variables.tf
MattiaSarti, line 17: thought: I am a little concerned with this variable naming, because we have other variables with similar names for other components (e.g., resource dispatcher) where profile means Kubeflow Profiles instead of testing/production mode

terraform-refactoring/components/data-kubeflow-integrator/variables.tf
```diff
@@ -0,0 +1,108 @@
...
+# See LICENSE file for licensing details.
+
+variable "model_uuid" {
+  description = "Reference to an existing model uuid."
+  type        = string
+  nullable    = false
+}
+
+variable "profile" {
+  description = "Name of profiles to apply this to"
```
MattiaSarti: nitpick: singular or plural?

MattiaSarti (changes requested):
todo: I'd like someone more familiar than me with Spark and Data Integrator to take a look too

terraform-refactoring/tests/conftest.py
deusebio: Not sure I follow. The user can parametrize the bucket name in the solution, however here we specify a given name just in the context of the tests. For the tests, I don't think it is needed to let the user parametrize the bucket name, no?

terraform-refactoring/tests/conftest.py
deusebio: This is not a requirement. These inputs are provided via tfvars. But for the context of the tests, we export those env variables that we load using the `load_dotenv`, and then we load them using `os.getenviron`. 

This is a pattern that has been using in the very same context to inject S3 credentials in spark, see [here](https://github.com/canonical/spark-k8s-bundle/blob/main/python/tests/integration/conftest.py#L56), and this should play nicely also when we migrate our testing to spread

terraform-refactoring/components/observability/integrations.tf
deusebio: I don't think there is such a thing. The otel_collector integrations would only be two: see [here](https://github.com/canonical/charmed-kubeflow-solutions/blob/main/modules/kubeflow/cos_configuration.tf#L590) and also [this comment](https://github.com/canonical/charmed-kubeflow-solutions/pull/189#discussion_r3430584610)

terraform-refactoring/products/kubeflow/main.tf
deusebio: uhm, that's good point. Would you have any suggestion as an alternative?

.github/workflows/deploy-to-k8s.yaml
deusebio: Why don't we want to allow specifying the risk for ambient? I mean how can we then progress the testing through the risk if we don't allow specifying the risk from edge to stable?

It also feels strange that - although we set in `dependencies` a risk level - when using ambient, we just override this

terraform-refactoring/tests/conftest.py
MattiaSarti: The fact is that those "tests" are not actual tests but the de-facto way to deploy our solutions with sensible defaults (without specifying all Terraform variables via `terraform apply` explicitly), and we, developers/solutions' users, often rely on a simple `tox` command to deploy everything in the same way as in our CI

.github/workflows/deploy-to-k8s.yaml
deusebio: This is copy-and-paste from [here](https://github.com/canonical/spark-k8s-bundle/blob/main/.github/actions/run-test/action.yaml#L204). I believe this is a good suggestion, but I'd rather stick to this for consistency honestly

terraform-refactoring/products/kubeflow/variables.tf
deusebio: uhm, yeh, that's good point, that in KF, the term "profile" has a different meaning, but it would also be nice to be consistent across products (hence the spec that I linked). 

Personally I would feel that so long as we have good documentation (meaning the input variable has a description), it should not be a problem.

terraform-refactoring/components/data-kubeflow-integrator/variables.tf
deusebio: Oh, yes! It should be singular! I'll fix this

terraform-refactoring/tests/conftest.py
MattiaSarti: Okay, it makes sense

terraform-refactoring/components/observability/integrations.tf
MattiaSarti: Okay, I didn't understand that that comment was about that third integration, so it was a mistake to have the third one there in the first place

.github/workflows/deploy-to-k8s.yaml
MattiaSarti: It was intentionally set that way because ambient mode was only available on edge risk. Do we not support it all the way up to stable risk? If so, I am fine with having it follow the set risk

.github/workflows/deploy-to-k8s.yaml
MattiaSarti: Sure, no problem!

terraform-refactoring/tests/conftest.py
deusebio: > de-facto way to deploy our solutions with sensible defaults (without specifying all Terraform variables via terraform apply explicitly)

But the sensible defaults are provided as defaults of the tf. But look, if you think it can be useful to have an env variable here as well, I can address this. I generally don't use the tox for deployment, only for UATs. But if you are using this also for deploying, that's fine to make it more usable.

terraform-refactoring/products/kubeflow/variables.tf
MattiaSarti: Okay, I'm fine with being coherent with the spec as well, if we document it well enough to disambiguate

terraform-refactoring/products/kubeflow/main.tf
MattiaSarti: Verbose as I tend to be, I would come up with something as `data_integrator_integrations`, but that's ugly to say the least... what about `data_integrations`? Or `external_integrations`, as in the comment above the variable?

.github/workflows/deploy-to-k8s.yaml
dariofaccin: I do think it's safe to remove the override here: it was a temporary workaround for ambient not being available in any other risk than `edge`.

terraform-refactoring/components/observability/integrations.tf
dariofaccin: Yes, this was overlooked by me when initially adding the COS integration. Thanks Enrico for catching and fixing it.

terraform-refactoring/products/kubeflow/main.tf
dariofaccin: This was confusing also to me. I agree that `external_integrations` might be more reasonable.

terraform-refactoring/tests/conftest.py
dariofaccin: I have almost never used `tox` to deploy the solution, exception made when I was trying to simulate the CI.
For local deployment I usually do `terraform apply -var-file="variables.tfvars"`.

MattiaSarti: Okay, then no need to update this, thanks anyway, @deusebio

deusebio: Well, I feel it is so little effort that if you are the only one using this, it can still be useful to have it :D I have added this in the review

terraform-refactoring/products/kubeflow/main.tf
deusebio: Done!

.github/workflows/deploy-to-k8s.yaml
deusebio: Resolving with no changes then!

terraform-refactoring/products/kubeflow/variables.tf
deusebio: Added extra documentation in the variable description

terraform-refactoring/components/data-kubeflow-integrator/variables.tf
deusebio: Done!

.github/workflows/deploy-to-k8s.yaml
deusebio: Resolving then!

terraform-refactoring/components/observability/integrations.tf
deusebio: No problem! Glad I could help!

dariofaccin (approved):

terraform-refactoring/products/kubeflow/main.tf
deusebio: moved the reference to the feature branch of the Spark TF refactoring.

terraform-refactoring/products/kubeflow/variables.tf
deusebio: Discussed, so resolving this!

terraform-refactoring/tests/conftest.py
MattiaSarti: Thanks, then!

MattiaSarti (approved):

PR #195, opened 2026-06-08
[PRA-7] feature: add Kubeflow - Spark integration Terraform bundle (#102)

deusebio:
Backporting https://github.com/canonical/charmed-kubeflow-solutions/pull/102 to main (since it originally landed on track/1.10)


modules/kubeflow-spark/main.tf
```diff
@@ -0,0 +1,111 @@
...
+  trust       = true
+  constraints = "arch=amd64"
+}
+
+resource "juju_application" "kubeflow_integrator" {
+  model = module.kubeflow.model
+  name  = "kubeflow-integrator"
+  charm {
+    name    = "data-kubeflow-integrator"
+    channel = "1/${var.risk}"
```
deusebio, line 102: **highlight** there is no `latest/edge` in kubeflow-data-integrator. I believe we should structure the common pattern that `main` also release to `latest`

deusebio: Tried to raise a PR to address this here: https://github.com/canonical/data-kubeflow-integrator/pull/32

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -36,10 +37,14 @@ jobs:
...
-            uats_branch: main
             uats:
               env: kubeflow-remote
-              extra-args: --bundle "file:assets/versions-ambient.yaml"
+              extra-args: --include-ambient-tests
+          - module: kubeflow-spark
+            tf-vars-file: examples/tfvars.json
+            uats:
+              env: spark-remote
+              extra-args: --test-image ghcr.io/canonical/charmed-spark-jupyterlab:3.5-22.04_edge@sha256:72a6e89985e35e0920fb40c063b3287425760ebf823b129a87143d5ec0e99af7 --bundle ""
```
misohu: question: If i understand correctly this image wont be changing with each version right ? We will most likely reuse it for future releases right ?

modules/kubeflow-spark/tests/test_deployment.py
misohu, line 1: suggestion (non-blocking): This file is nice but some of these functions are shared accross all the other tests and we are just copy pasting them .... maybe its worth to open an tech debt issue to refactor in the fututre

dariofaccin (approved):
LGTM.

modules/kubeflow-spark/tests/test_deployment.py
dariofaccin: @misohu I think we can handle this one when migrating to the new Terraform modules.

deusebio: Yes. I fully agree that it would be nice to centralize these, but this will require a substantial refactoring, which as @dariofaccin highlights, I feel it would not be worth here given the new tf refactoring we have (which it should also help to just have one solution and not have multiple modules for each combination)

.github/workflows/deploy-to-k8s.yaml
deusebio: Uhm, that's very good point. This was already the case in the first PR, but I agree that this will require substantial effort to maintain. Let me change this!

We can probably just use `edge` and `stable` depending whether we are in a stable branch, or in main. I'll update this in the current PRs and raise a new PR on the track/1.10

misohu (approved):

PR #215, opened 2026-07-06
bug: improve UATs stability

deusebio:
Improving UATs stability by:

1. Ensuring network availability when test-kubeflow pod is started, see https://github.com/canonical/charmed-kubeflow-uats/pull/280
2. Preloading images before running UATs


.github/workflows/deploy-to-k8s.yaml
```diff
@@ -36,9 +37,9 @@ jobs:
               env: kubeflow-remote
           - product: kubeflow
             service-mesh-type: ambient
-            uats_branch: main
             uats:
               env: kubeflow-remote
+              extra-args: --include-ambient-tests
```
MattiaSarti: suggestion (non-blocking): this will add the flag only to the matrix combination with product `kubeflow` and service-mesh type `ambient`, while we may also want any matrix combinations with products in ambient mode to achieve the same - I know most of the failing UATs were for Kubeflow itself, like Katib's, but it would be nice to include ambient tests for any product in ambient mode

&&

thought: it would be nice to reuse `service-mesh-type` to then decide when to include this flag, but I don't know how to technically achieve it - if it's not possible, we'll have to define explicitly all matrix combinations

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -181,13 +180,82 @@ jobs:
...
+          done < $IMAGE_LIST
+
+      - name: Run UATs
+        run: |
+          cd ~/charmed-kubeflow-uats	  
           tox -e ${{ matrix.uats.env }} -- ${{ matrix.uats.extra-args }} --bundle=""
 
+      - name: Setup tmate session
+        if: failure()
+        uses: canonical/action-tmate@main
```
dariofaccin, line 249: suggestion(non-blocking): remove.

.github/dependencies.yaml
```diff
@@ -1,5 +1,5 @@
 CKF_BRANCH: "track/1.11"
 K8S_VERSION: "1.32"
 JUJU_VERSION: "3.6"
-UATS_BRANCH: "track/1.11"
+UATS_BRANCH: "wip-improve-uats-stability-1.11"
```
dariofaccin: suggestion: remember to revert this change.

dariofaccin (approved):

MattiaSarti (approved):

.github/workflows/deploy-to-k8s.yaml
dariofaccin: @MattiaSarti Bonding `--include-ambient-tests` to `service-mesh-type` is simple, as we can re-use the `if` clause of L170.

MattiaSarti: Yes, we can always do that if there's no way not to restructure the command `tox -e ${{ matrix.uats.env }} -- ${{ matrix.uats.extra-args }} --bundle=""`

deusebio: How about the changes here: https://github.com/canonical/charmed-kubeflow-solutions/pull/219/changes

I'm basically leveraging on the `include` capabilities of Github actions. Although I find it elegant to address the concern of Mattia, there are a couple of shortcomings (which are not critical to address the point here):
1. We can't use the nested structure under `uats` as github can't merge dictionaries, so I flattened the key to allow composition (with uats-env coming from the include of the product, and `uats-extra-args` coming from the integration with the service mesh)
2. We can't combine multiple uats-extra-args coming both from product (e.g. spark) and service-mesh. This is not critical right now given spark does not have ambient

We could start with the code in the PR above for the time being. But also bear in mind that I'm also planning to provide some auto-inference capabilities that will turn on tests based on which charms are deployed. E.g. if istio-beacon-k8s is deployed, then ambient tests are triggered. However to enable this, I need to first land the Jubilant changes that will make this possible

MattiaSarti: Sure, this all makes sense

deusebio: merged the PR above

.github/dependencies.yaml
deusebio: done!

.github/workflows/deploy-to-k8s.yaml
deusebio: I would leave it as it allows debugging only when failure happens. I believe this can be useful to at least debug, given the limited debugging ability of UATs. Given it is non-blocking, I'd rather leave it in

PR #218, opened 2026-07-13
feat: add multi-tenancy support for KFP

dariofaccin:
## Summary

This branch adds multi-tenancy support for Kubeflow Pipelines (KFP) and reworks
object storage across the product. KFP's `kfp-profile-controller` is wired to the
`resource-dispatcher` through the new `config-maps` and `secrets` endpoints, and
`resource-dispatcher` becomes a first-class, always-deployed component.

Object storage is moved off `minio` and onto a shared `s3-integrator`
(`s3-integrator-global`) exposing the `s3-credentials` interface. KFP, MLflow and
KServe now consume object storage from this shared integrator, and the `minio`
module is dropped entirely. Tests and CI are updated to provision the required S3
users/buckets.

### KFP ↔ resource-dispatcher integration

- Exposed additional endpoints on the `resource-dispatcher` charm module outputs:
  `config_maps`, `provide_cmr_mesh` (provides) and `require_cmr_mesh` (requires).
- Added `config_maps` and `secrets` input variables to the `kfp` component
  (same endpoint/offer object shape and validation as the other integration
  inputs).
- Added `kfp_profile_controller_config_maps` and `kfp_profile_controller_secrets`
  integrations so `kfp-profile-controller` consumes `config-maps` and `secrets`
  from `resource-dispatcher`.
- Product `kubeflow`: the `kfp` module now `depends_on` `module.resource_dispatcher`
  and receives the `config_maps` and `secrets` endpoints.

### resource-dispatcher always deployed

- Removed the `count` guard on `module.resource_dispatcher` so it is always
  deployed, and dropped the `[0]` index from every reference across the product
  (`mlflow`, `feast`, `kserve`, `external_integrations`, and the service-mesh
  integration).
- The `resource_dispatcher_service_mesh` integration is now created whenever the
  service mesh type is `ambient`, independent of `enable_mlflow`/`enable_feast`.

### Object storage via a shared S3 integrator

- Added a shared `s3-integrator-global` application (Juju secret + secret access),
  deployed whenever `enable_kfp` or `enable_mlflow` is set. Configured through
  `s3_*_global` variables (`s3_secret_key_global`, `s3_access_key_global`,
  `s3_endpoint_global`, `s3_bucket_global`, `s3_config_global`,
  `s3_revision_global`).
- KFP: removed the `object_storage` (minio) variable and integrations; the
  `argo-controller`, `kfp-api`, `kfp-profile-controller` and `kfp-ui`
  applications now consume the `s3-credentials` interface via a new
  `s3_credentials` variable.
- MLflow: removed the `object_storage` (minio) variable and integration;
  `mlflow-server` now consumes `s3-credentials` via a new `s3_credentials`
  variable.
- KServe: the `kserve-controller` object-storage relation (used to read MLflow
  model artifacts) is repointed to the shared `s3-integrator` over the
  `s3-credentials` interface.

### Drop the minio module

- Removed the `module "minio"` block and its wiring: the `deploy_minio` and
  `minio_channel` locals, the `minio_*` product variables, the
  `minio_service_mesh` integration, and `module.minio` from the observability
  `depends_on`.
- Removed the minio observability wiring: the `minio_grafana_dashboard` /
  `minio_metrics_endpoint` inputs in the product and the corresponding variables
  and integrations in the `observability` component.

### Spark S3 variables renamed

- Renamed the existing Spark S3 resources and variables from `s3_*` to `s3_*_spark`
  (and the app to `s3-integrator-spark`) to disambiguate them from the shared
  integrator.

### Tests and CI

- `tests/conftest.py`: added the `setup_s3_integrator_global` fixture (wiring the
  `s3_*_global` variables) and updated the Spark fixture / `tf_vars`; dropped the
  now-removed `minio_storage_size` var.
- `tests/tox.ini`: updated `passenv` with the Spark / global / override S3
  environment variables.
- `.github/workflows/deploy-to-k8s.yaml`: the object-storage setup step now runs
  for all products and provisions MicroCeph users/buckets for the
  spark / global / override S3 tenancy scenarios.

### Misc

- Switched the `istio-ambient` module source in the product from a git ref to a
  local path.
- Regenerated the affected `README.md` files via `terraform-docs`.

## Testing

- `terraform fmt` passes on the changed modules.
- `terraform validate` passes for the `kfp`, `mlflow`, `kserve` and
  `observability` components and for the `kubeflow` product.


.github/workflows/deploy-to-k8s.yaml
```diff
@@ -126,7 +125,10 @@ jobs:
                 1: sudo microceph cluster bootstrap
                 2: sudo microceph disk add loop,1G,3
                 3: sudo microceph enable rgw
-                4: sudo microceph.radosgw-admin user create --uid test --display-name test --access-key=foo --secret-key=bar
+                4: sudo microceph.radosgw-admin user create --uid test-spark --display-name test-spark --access-key=foo --secret-key=bar
+                5: sudo microceph.radosgw-admin user create --uid test-global --display-name test-global --access-key=baz --secret-key=qux
+                6: sudo microceph.radosgw-admin user create --uid test-override --display-name test-override --access-key=grault --secret-key=corge
```
mvlassis, line 152: note: I had to search for these, the terms `qux`, `grault`, and `corge` are common metasyntantic variables for placeholders like `foo` and `bar`

terraform/tests/conftest.py
```diff
@@ -100,22 +100,44 @@ def enable_spark(request) -> list[str]:
...
+            f"s3_endpoint_spark={os.environ['S3_SERVER_URL_SPARK']}",
         ]
         print(f"Extra args for Spark deployment: {extra_args}")
         return extra_args
     return []
 
 
+@pytest.fixture(scope="module")
+def setup_s3_integrator_kfp() -> list[str]:
+    """Terraform module customization for Spark deployment."""
```
mvlassis: nitpick: We probably mean the `KFP` deployment here, right? It's using the KFP variables
```suggestion
    """Terraform module customization for KFP deployment."""
```

terraform/tests/conftest.py
```diff
@@ -100,22 +100,44 @@ def enable_spark(request) -> list[str]:
...
+        "-var",
+        f"s3_bucket_kfp={os.environ['S3_BUCKET_KFP_GLOBAL']}",
+        "-var",
+        f"s3_secret_key_kfp={os.environ['S3_SECRET_KEY_GLOBAL']}",
+        "-var",
+        f"s3_access_key_kfp={os.environ['S3_ACCESS_KEY_GLOBAL']}",
+        "-var",
+        f"s3_endpoint_kfp={os.environ['S3_SERVER_URL_GLOBAL']}",
+    ]
+    print(f"Args for s3 KFP integration: {args}")
```
mvlassis: nitpick: Capitalization, S3 always appears as `S3`
```suggestion
    print(f"Args for S3 KFP integration: {args}")
```

terraform/products/kubeflow/main.tf
```diff
@@ -249,9 +249,54 @@ module "katib" {
...
+  value = {
+    secret-key = var.s3_secret_key_kfp
+    access-key = var.s3_access_key_kfp
+  }
+  info = "This is the access key and secret key for the S3 storage"
+}
+
+module "s3_kfp" {
+  depends_on = [juju_model.kubeflow, juju_secret.s3_secret_kfp]
+  count      = var.enable_spark ? 1 : 0
```
mvlassis: question(blocking): This is incorrect, right? We want the S3 integrator for KFP to deploy based on whether `enable_kfp` is true. Otherwise, enabling Spark is *required* for deploying the module
```suggestion
  count      = var.enable_kfp ? 1 : 0
```

mvlassis (commented):
Tested on a VM and the deployment worked correctly (after applying the suggestion I have left regarding the kfp secret:
```
terraform apply -var create_model=false -var s3_access_key_kfp=foo -var s3_secret_key_kfp=bar -var s3_endpoint_kfp=http://10.31.200.197 -var model_uuid=3943ff9e-72a7-4cdd-8dfc-9b0cfe32ad20
```
## Notes
- We should probably also pass the [tls-ca-chain](https://github.com/canonical/object-storage-integrator/tree/main/s3#further-configuration) config option to the s3-integrator. Although `kfp` and `argo` don't use it currently, we may add support in the future, and `mlflow-operator` already can use it.
- We should add the `s3-credentials` integrations to `kfp-ui`, `kfp-profile-controller`, `kfp-api`, and `argo-controller`.
- Although not strictly relevant for this multi-tenancy in KFP, we could add the `s3-credentials` integrations to `mlflow-operator` and `kserve-operator`.
- We *shouldn't* drop the object_storage variable for `mlflow-operator`, and in all other charms as of now, for compatibility reasons

terraform/tests/conftest.py
dariofaccin: Renamed the function since this S3 integrator is the global one (needed for KFP, MLFlow and KServe).

terraform/products/kubeflow/main.tf
dariofaccin: It was completely wrong. Now we deploy `s3_global` when either KFP or MLFlow are required.

terraform/products/kubeflow/integrations.tf
```diff
@@ -57,6 +58,23 @@ resource "juju_integration" "kserve_controller_object_storage" {
   }
 }
 
+# kserve-controller s3-credentials integration (s3-integrator-global:s3-credentials -> kserve-controller)
+# Only deployed with the 's3' object storage mode and when MLflow is enabled, since kserve uses this storage to read MLflow model artifacts
+resource "juju_integration" "kserve_controller_s3_credentials" {
```
mvlassis, line 211: I think relying on mlflow here is strictly not correct. We can have a `kserve-controller:s3-credentials` integration without having an `mlflow-operator` deployed. The integration is used for potentially deploying models from the S3 storage, and they can be stored there without `mlflow` (via a simple `cp` for example).

As a result, `mlflow` shouldn't be a requirement

dariofaccin: Addressed.

terraform/components/kfp/integrations.tf
```diff
@@ -37,6 +55,40 @@ resource "juju_integration" "kfp_api_mysql_database" {
   }
 }
 
+# S3 credentials integrations (s3-integrator:s3-credentials -> KFP apps)
+
+resource "juju_integration" "kfp_api_s3_credentials" {
```
deusebio, line 60: **question** what is this for? this relation does not appear in the spec KF137. Also, for what are these credentials being used? is this limiting to some extent the multi-tenancy of KFP since the API would just use one particular bucket?

It would be good to clarify this in the spec as well since this integration was somewhat unexpected for me @mvlassis

terraform/components/kfp/integrations.tf
```diff
@@ -71,6 +123,56 @@ resource "juju_integration" "kfp_profile_controller_object_storage" {
...
+  }
+
+  application {
+    name      = var.secrets.kind == "endpoint" ? var.secrets.name : null
+    endpoint  = var.secrets.kind == "endpoint" ? var.secrets.endpoint : null
+    offer_url = var.secrets.kind == "offer" ? var.secrets.url : null
+  }
+}
+
+resource "juju_integration" "kfp_ui_s3_credentials" {
```
deusebio, line 160: **thought** as before, I was somewhat not expecting this. Let's just make sure this does not limit/prevent the full multi-tenancy of the deployment

terraform/components/kfp/outputs.tf
```diff
@@ -55,6 +55,10 @@ output "provides" {
 output "requires" {
   description = "Map of endpoints required by this component from other components (inbound relations)"
   value = {
+    kfp_api_s3_credentials = {
```
deusebio, line 58: **niptick** since these endpoints are not quite consumed externally at this point (since they are already wired up in the module itself with input variable) they may not be required to be exposed outside

deusebio (commented):
The code changes look good and they were *almost* what I was expecting. I was just surprised about a couple of relations I'd like to get a bit more clarification, just to understand how those credentials are centrally used. 

But apart from that, I believe the code is already quite good to be merged.

terraform/components/kfp/integrations.tf
mvlassis: This specific relation doesn't affect multi-tenancy. I created https://github.com/canonical/kfp-operators/issues/963 to track why it's used, the gist of it is that it is used for:
- Checking and creating the `mlpipeline` (or specified name) bucket if it doesn't exist
- Having a fallback if the backend cannot find the pipeline IR (intermediate representation) in the MySQL storage.

The presence of this integration *doesn't* affect multi-tenancy, and as of now *has* to be present, otherwise the backend executable will fail to check the bucket in the specified object storage, and exit

terraform/components/kfp/integrations.tf
mvlassis: This is similar to the `kfp-api` case, we are essentially replacing the already-existing `kfp-ui:object-storage` interface which is mandatory.

mvlassis (approved):
Tried on a Testflinger instance:
- Deploy the solution from this branch
- Clone the UATS repo and checkout to the `feat/KF-8788-multi-tenancy-kfp` branch
- Export the `S3` environment variables

And the tests run successfully

terraform/components/kfp/integrations.tf
deusebio: All right! Perfect! It is a bit *suboptimtal* to have an effectively useless relation, only there for doing the checks (less worried about the fallback since that should NEVER happen), but ok, we can't change the upstream component

terraform/components/kfp/integrations.tf
deusebio: All right! Thanks! As long as it does not prevent the multi-tenancy proper user experience, that's fine.

deusebio (approved):
@dariofaccin If you don't manage to get this to have green CI by today, I can take this over next week to stabilize the UATs on CI. I've become a pro on this :P . The changes here are good (so I'll approve), but I would like to merge this once the CI is green without too much flakiness, since otherwise it will become impossible to develop on `main` cleanly

dariofaccin:
Closing as superseded and merged in #246.

PR #220, opened 2026-07-14
feat: IAM integration

dariofaccin:
## Summary

Introduces an **ambient-mesh, Canonical Identity Platform (IAM)–based** authentication architecture for the Terraform refactoring, replacing the legacy Dex/OIDC `auth` stack on the primary ambient path. This implements the [IAM + Istio Decoupling Refactor spec](terraform-refactoring/specs/iam-istio-auth-refactor.md).

The solution now splits responsibilities across **three Juju models** — `istio-system` (Istio control plane), `iam` (Identity Platform), and `kubeflow` (apps + gateways + IAM auth charms) — composed by a single top-level deployment that wires the cross-model offers.

A parallel **ambient + Dex** path (`istio-ambient-dex`) is retained for the Dex/OIDC-based authentication variant.

## What's included

**New charm modules** (`terraform-refactoring/charms/`)
- `istio-k8s` — ambient control plane, deployed in its own `istio-system` model
- `istio-ingress-k8s` — gateway charm, instantiated twice (`-ui` and `-m2m`)
- `istio-beacon-k8s` — ambient mesh beacon
- `oauth2-proxy-k8s`, `request-authentication-configurator`, `github-profiles-automator` — IAM auth charms

**New `iam` product** (`terraform-refactoring/products/iam/`)
- Wraps the Canonical Identity Platform (Hydra / Kratos / Login UI) plus in-model deps (`postgresql-k8s`, `traefik-k8s`, `self-signed-certificates`)
- Reuses a **vendored + patched** [`iam-bundle-integration`](terraform-refactoring/vendor/iam-bundle-integration) module (pinned to `v1.1.1`, juju provider constraint relaxed to `~> 1.0` for juju 1.x compatibility)
- Exposes `oauth_offer_url` (Hydra) and `send_ca_cert_offer_url` cross-model

**`kubeflow` product updates**
- Adds the three IAM auth charms and the two ambient gateways to the ambient path
- New integrations, locals, and variables to wire the cross-model offers (`istio_ingress_config`, `oauth`)
- Reworked `main.tf`/`locals.tf` count logic and configuration surface

**Istio ambient components**
- `istio-ambient` — reworked applications/integrations/outputs/variables to consume the control-plane ingress-config offer and support the dual-gateway (UI/M2M) topology
- `istio-ambient-dex` — new component for the ambient + Dex/OIDC authentication variant

**Top-level deployment** (`terraform-refactoring/tests/kubeflow-ambient-iam/`)
- Single Terraform root composing the `istio-system` model + `iam` + `kubeflow` products with all cross-model wiring

## Cross-model topology

- `istio-k8s:istio-ingress-config` — offered from `istio-system`, consumed by both gateways in `kubeflow`
- `hydra:oauth` — offered from `iam`, consumed by `oauth2-proxy` and `request-authentication-configurator` in `kubeflow`
- `self-signed-certificates:send-ca-cert` — offered from `iam-core`, consumed cross-model (e.g. `istio-k8s:jwks-ca-cert`, `oauth2-proxy:receive-ca-cert`)

## Notes

- Ambient Istio charms pinned to `dev/edge`; IAM charms to `latest/edge` (only channels exposing the required endpoints/config).
- `terraform validate` passes for all modules and the deployment.

## Testing

- [ ] `terraform validate` across all new/changed modules
- [ ] End-to-end `kubeflow-ambient-iam` deployment


terraform/tests/kubeflow-ambient-iam/main.tf
NohaIhab, line 1: suggestion: Shall we move it out of the `tests` dir? It should be reachable like the other solutions

scripts/tox.ini
```diff
@@ -60,6 +60,12 @@ commands =
     terraform -chdir={[vars]tf_modules_path}tests/kubeflow-cos fmt -recursive -check -diff
     tflint --recursive --chdir={[vars]tf_modules_path}tests/kubeflow-cos
 
+[testenv:lint-kubeflow-ambient-iam]
+description = Check Terraform code against coding style standards
+commands =
+    terraform -chdir={[vars]tf_modules_path}tests/kubeflow-ambient-iam fmt -recursive -check -diff
+    tflint --recursive --chdir={[vars]tf_modules_path}tests/kubeflow-ambient-iam
+
```
NohaIhab, line 68: based on my previous comment, we can group this with the other lint envs

terraform/charms/github-profiles-automator/main.tf
NohaIhab, line 1: question: why are we not using the tf module from the github-profiles-automator repo?

terraform-refactoring/charms/istio-beacon-k8s/main.tf
NohaIhab, line 1: suggestion: they already have a module we can use in https://github.com/canonical/istio-beacon-k8s-operator/tree/main/terraform

terraform-refactoring/charms/istio-ingress-k8s/main.tf
NohaIhab, line 1: suggestion(same as above): they already have a module in https://github.com/canonical/istio-ingress-k8s-operator/tree/main/terraform

terraform-refactoring/charms/istio-k8s/main.tf
NohaIhab, line 1: suggestion(same as above): they already have a module in https://github.com/canonical/istio-k8s-operator/tree/main/terraform

terraform/charms/oauth2-proxy-k8s/integrations.tf
```diff
@@ -0,0 +1,40 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+# oauth2-proxy -> hydra (oauth2-proxy:oauth -> hydra:oauth). Supports a
+# same-model endpoint or a cross-model offer (Hydra in the iam model).
```
NohaIhab, line 5: note: they will always be related cross-model because oauth2-proxy gets deployed in `kubeflow` model

terraform/charms/oauth2-proxy-k8s/main.tf
NohaIhab, line 1: suggestion(same as above): there's a tf module in https://github.com/canonical/oauth2-proxy-k8s-operator/tree/main/terraform

terraform/charms/request-authentication-configurator/main.tf
NohaIhab, line 1: question: why are we not using the tf module from the `request-authentication-configurator` repo?

terraform-refactoring/charms/github-profiles-automator/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Providers
+
+| Name | Version |
+| ---- | ------- |
+| <a name="provider_juju"></a> [juju](#provider\_juju) | >= 1.1.1 |
+
+## Modules
+
+No modules.
```
NohaIhab: not sure what you mean by this, it's confusing, also present in other readmes

terraform-refactoring/components/istio-ambient/README.md
```diff
@@ -8,30 +8,29 @@
 
 ## Providers
 
-| Name | Version |
-| ---- | ------- |
-| <a name="provider_juju"></a> [juju](#provider\_juju) | >= 1.1.1 |
+No providers.
```
NohaIhab: again, why is this omitted?

terraform/components/istio-ambient/variables.tf
```diff
@@ -7,22 +7,25 @@ variable "model_uuid" {
...
+    kind     = string
+    name     = optional(string, null)
+    endpoint = optional(string, null)
+    url      = optional(string, null)
   })
-  default = {}
+  default = null
 }
 
 variable "istio_ingress_k8s" {
```
NohaIhab, line 27: suggestion(blocking): this should be renamed based on the usage now

terraform/products/kubeflow/variables.tf
```diff
@@ -330,13 +330,13 @@ variable "kfp_viz_config" {
 # Istio Component Applications
 
 variable "service_mesh_type" {
```
NohaIhab, line 332: nit: update the variable name to be representative

terraform/products/kubeflow/variables.tf
```diff
@@ -367,19 +367,19 @@ variable "istio_ingressgateway_config" {
 # Ambient Component Applications
 
 variable "istio_k8s_revision" {
-  description = "Revision of the istio-k8s application"
+  description = "Revision of the istio-k8s control-plane charm (ambient-dex only; istio-k8s runs in-model there)."
```
NohaIhab, line 386: suggestion(non-blocking): it's not a requirement for istio-k8s to run in the kubeflow model with dex, actually it's best practice to have in separate model i.e. `istio-system`. It can be beyond the scope of this PR but ideally it should be always in a separate namespace.

terraform/tests/kubeflow-ambient-iam/main.tf
```diff
@@ -0,0 +1,134 @@
...
+
+locals {
+  istio_system_model_uuid = var.create_istio_system_model ? juju_model.istio_system[0].uuid : var.istio_system_model_uuid
+
+  # Merge the top-level external hostname vars into the per-component config
+  # maps (only when set, so an unset hostname never injects an empty value).
+  istio_ingress_k8s_ui_config = merge(
+    var.istio_ingress_k8s_ui_config,
+    var.external_ui_hostname != null ? { external_hostname = var.external_ui_hostname } : {}
+  )
```
NohaIhab, line 24: question: IIUC if someone sets both, the external hostname var will supersede?

terraform/tests/kubeflow-ambient-iam/outputs.tf
NohaIhab, line 1: suggestion(non-blocking): I think it's useful to output the hostnames as well, not sure if it makes sense since they can be set by users

NohaIhab (commented):
Thanks @dariofaccin for supporting on Identity, commented open for discussion

terraform/tests/kubeflow-ambient-iam/main.tf
dariofaccin: Also the `kubeflow-cos` deployment is defined under `tests`. Let's see what @deusebio thinks.

terraform/charms/github-profiles-automator/main.tf
dariofaccin: That module, and in general all our charm modules, is still using Juju provider `0.x` which does not support `model_uuid`. To migrate it, we need first to deprecate the Kubeflow modules in this repo.

terraform/charms/request-authentication-configurator/main.tf
dariofaccin: See comment for the other charm.

terraform/charms/oauth2-proxy-k8s/main.tf
dariofaccin: See comment for the other charm (`version = ~> 1.0`).

terraform-refactoring/charms/istio-k8s/main.tf
dariofaccin: See comment for the other charm (`version = ~> 1.0`).

terraform/charms/oauth2-proxy-k8s/integrations.tf
dariofaccin: Since this is a simple charm, we must remain agnostic on the integration "type" (in-model or cross-model).

terraform-refactoring/charms/github-profiles-automator/README.md
dariofaccin: Removed.

terraform-refactoring/components/istio-ambient/README.md
dariofaccin: Fixed.

terraform/products/kubeflow/variables.tf
dariofaccin: I think we can "easily" address this by moving the creating (or passing an existing one as a var) of the `istio-system` model inside the Kubeflow product model. Given the size of this PR I'd prefer to keep the legacy behaviour.

terraform/tests/kubeflow-ambient-iam/main.tf
dariofaccin: According to the `merge` documentation[^1]:
> If more than one given map or object defines the same key or attribute, then the one that is later in the argument sequence takes precedence.

[^1]: https://developer.hashicorp.com/terraform/language/functions/merge

terraform/tests/kubeflow-ambient-iam/main.tf
NohaIhab: I'm not sure why `kubeflow-cos` is under `tests/` either, if we want to group integrations with other solutions we can call it something suitable for that.

terraform/products/kubeflow/variables.tf
NohaIhab: ok, I filed #225 to track this

scripts/tox.ini
NohaIhab: keeping this open until we resolve https://github.com/canonical/charmed-kubeflow-solutions/pull/220#discussion_r3577373760

terraform/charms/oauth2-proxy-k8s/main.tf
NohaIhab: got it, I'll resolve this one because it's not owned by our team

terraform/products/kubeflow/variables.tf
NohaIhab: what I meant is, it's now about the service mesh and auth type too

terraform/tests/kubeflow-ambient-iam/main.tf
deusebio: I believe these are meant to be "Deployments" modules used as **examples**. The modules here are not to be used by other teams, but rather to be used by us to test our module, hence why I find "tests" naming ok. Another deployment will need to integrate the various product modules, similarly to what we do here, but not reference these modules directly. 

Based on this, I believe it is fine to leave them in a folder called `tests`. Maybe my preference would be to move this folder at root level, considering the "terraform" folder a sort of "src" folder, and nudging away other teams to consume these folders that can be found alongside the others. But this is a minor and we can do it later if we agree that this would be beneficial. For now, fine to keep it where it currently is

dariofaccin: Yes, I think we can have an `examples` folder with these deployments.

NohaIhab: I'm good with `examples` too, and it's ok to keep it as is now. Thanks @deusebio.

terraform/products/kubeflow/variables.tf
NohaIhab: thank you for splitting into 2 variables

terraform/tests/kubeflow-ambient-iam/outputs.tf
dariofaccin: I don't have a strong opinion. These being inputs, I don't see a point of exposing them as outputs.
I am more interested in understanding if they should have a default, or improve validation (avoiding a `null` input, for example).

NohaIhab: For the record we discussed async and agreed to enforce the hostname inputs and validate them, thanks for pushing the change in https://github.com/canonical/charmed-kubeflow-solutions/pull/220/changes/a359254be25759c6e352ed9af46db8a5a88cf877!

NohaIhab (approved):
Thank you @dariofaccin!

deusebio (approved):
The changes looks sane, and I believe the modifications are good enough to be merged in the context of this work (when you get the *reasonably* CI happy), although there are a few things will need to be addressed in the mid-term:

1. Removing the component module and the product module for IAM. These should live in the identity repositories. We can probably show them our structure and module and nudge an implementation that is similar (thus coherent with CC008) such that re-work on our side will be minimal

2. Refactor the istio components since these implementations are not quite Kubeflow specifics, even their integation with Identity. Ideally we should be consumer of those modules and we should not have components to do this, but rather possibly integrate those either at product level or at deployment level. This will require a bit of thoughts and discussions, but the current implementation with `istio-sidecar`, `isito-ambient` and `istio-ambient-dex`, does not seem the long term solution to this. I don't have any immediate solution or better proposal, as this will require discussion with service mesh/identity, but let's start to have this in the back of our minds

Please create some tickets on these such that we keep those in mind, and after the summer break we stream these discussions with the identity/service mesh teams

NohaIhab:
There is some flakiness with the CI, however the failures in the last commit were:
* pre-load images, there is a known issue tracked in [IS](https://warthogs.atlassian.net/browse/PFEREQ-26?atlOrigin=eyJpIjoiMWNlYTYyYjFmYzIyNDU0ZmJkNzMwZmJjMGYzOGY2MTIiLCJwIjoiaiJ9)
* microk8s setup
* UATs failing because of KFP stuck in `Running` before we merged https://github.com/canonical/charmed-kubeflow-uats/pull/290

I will merge to `main` and keep watching for flakiness while working on #235, and file issues where needed.

PR #223, opened 2026-07-14
chore: renaming terraform folder

deusebio:
Renaming the terraform folder to provide consistent UX with Spark K8s bundle, see https://github.com/canonical/spark-k8s-bundle


terraform/tests/kubeflow-cos/README.md
```diff
@@ -17,7 +17,7 @@ Prometheus metrics).
 
 ```hcl
 module "kubeflow_cos" {
-  source = "git::https://github.com/canonical/charmed-kubeflow-solutions//terraform-refactoring/products/kubeflow-cos?ref=main"
```
MattiaSarti, line 20: note: shouldn't we change this at a later point in time, if it's coming from `main`?

MattiaSarti (approved):

dariofaccin (approved):

terraform/tests/kubeflow-cos/README.md
deusebio: possibly, but this is only in the README.md, serving as an example. People should be familiar with the TF syntax, and they can update based on their need.

PR #224, opened 2026-07-16
Rebase and cleanup on main

deusebio:
Major changes:

* Removing modules with the old standard
* Updating CI files. In particular, When doing the rebase, I have noticed that the CI in the terraform/refactoring was actually not testing on MicroK8s. Testing on MicroK8s is not mandatory but a nice-to-have (see [here](https://discourse.canonical.com/t/charming-engineering-lt-notes-10th-june-2026/7602)). However, I noticed that current testing was actually not covering the PSS use-case, that was only implemented for MicroK8s. So I thought that it was worth keeping MicroK8s but only to also test PSS, that applies to side-car
* Updating KFP to 2.16 given it was the one we have in main and used in UATs on main


.github/workflows/deploy-to-k8s.yaml
```diff
@@ -19,6 +19,7 @@ on:
   schedule:
     - cron: "17 02 * * 1"
 
+
```
dariofaccin, line 22: nitpick: this new empty line is bugging me.

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -28,27 +28,21 @@ jobs:
...
+            uats-env: feast-remote
         pod-security-standards:
           - policy: privileged
             istio-cni-bin-dir: ""  # meaning Istio CNI disabled
             istio-cni-conf-dir: ""  # meaning Istio CNI disabled
           - policy: baseline
             istio-cni-bin-dir: /var/snap/microk8s/current/opt/cni/bin
             istio-cni-conf-dir: /var/snap/microk8s/current/args/cni-network
-
+
```
dariofaccin, line 46: nitpick: these whitespaces are bugging me.

dariofaccin (approved):
LGTM. Left a couple of nits.

.github/workflows/deploy-to-k8s.yaml
deusebio: Unless there will be major comments, I'll address there in the main PR to not retrigger the entire CI here for nipticks

.github/workflows/deploy-to-microk8s.yaml
deusebio: Unless there will be major comments, I'll address there in the main PR to not retrigger the entire CI here for nipticks

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -28,27 +28,21 @@ jobs:
...
-            uats:
-              env: feast-remote
-          - module: kubeflow-ambient
-            uats_branch: main
-            uats:
-              env: kubeflow-remote
-              extra-args:
+        product: 
+          - name: kubeflow
+            uats-env: kubeflow-remote
```
MattiaSarti, line 33: question: does this still default to Kubeflow in ambient mode, to be equivalent to what was there before?

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -137,28 +137,96 @@ jobs:
...
             git checkout '${{ env.CKF_BRANCH }}'
           fi
 
           CNI_ARGS="--istio-cni-bin-dir ${{ matrix.pod-security-standards.istio-cni-bin-dir }} --istio-cni-conf-dir ${{ matrix.pod-security-standards.istio-cni-conf-dir }}"
           RISK_ARG="--risk ${{ env.RISK }}"
-          PLATFORM_ARG=""
-          if [[ "${{ matrix.bundle.module }}" == "kubeflow-ambient" ]]; then
-            CNI_ARGS=""
-            RISK_ARG=""
-            PLATFORM_ARG="--istio-k8s-platform microk8s"
```
MattiaSarti, line 151: question: why are we no longer passing these ones for ambient mode?

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -137,28 +137,96 @@ jobs:
...
+              echo "::error::Logs:"
+              cat /tmp/import.out
+              exit 1
+            fi
+          done < $IMAGE_LIST
+
+      - name: Run UATs
+        run: |
+          cd ~/charmed-kubeflow-uats	  
+          tox -e ${{ matrix.product.uats-env }} -- --bundle="" --security-policy ${{ matrix.pod-security-standards.policy }} ${{ matrix.uats-extra-args }}
```
MattiaSarti, line 235: todo: we should remove `--bundle=""` in the final pull requests, later on

deusebio:
@MattiaSarti You comment is spot on, and I should have clarified this in the PR description (so I added the rationals below to the main description). 

When doing the rebase, I have noticed that the CI in the terraform/refactoring was actually not testing on `MicroK8s`. Testing on MicroK8s is not mandatory but a nice-to-have (see [here](https://discourse.canonical.com/t/charming-engineering-lt-notes-10th-june-2026/7602)). However, I noticed that current testing was actually not covering the PSS use-case, that was only implemented for MicroK8s. So I thought that it was worth keeping MicroK8s but only to also test PSS, that applies to side-car

I would be nice to actually test that when deploying with ambient, we require no higher privileges, but we can do this as a tech debt.

.github/workflows/deploy-to-microk8s.yaml
deusebio: I was discussing this with @NohaIhab earlier today. I believe it is fine to keep the bundle correctness disabled in main for the time being. I'm also about to raise a PR to improve this, as a follow up of the jubilant migration

.github/workflows/deploy-to-microk8s.yaml
deusebio: No, we are not testing on ambient mode on MicroK8s. See the comment [here](https://github.com/canonical/charmed-kubeflow-solutions/pull/224#issuecomment-5020562482) and the PR (updated) description

.github/workflows/deploy-to-microk8s.yaml
deusebio: No, I believe the default here is actually "sidecar", see https://github.com/canonical/charmed-kubeflow-solutions/blob/feat/terraform-refactor-main/terraform/products/kubeflow/variables.tf#L335

MattiaSarti (commented):
Thanks for clarifying, @deusebio!

> So I thought that it was worth keeping MicroK8s but only to also test PSS, that applies to side-car

This it not entirely accurate, as PSS do not apply only to sidacar mode. They can be set in both modes, irrespective of Istio's setup, and the previous setup encompassed testing PSS in both sidecar and ambient mode, so this way we would drop the testing of PSS in ambient mode that we used to always run before

deusebio:
@MattiaSarti I have added ambient tests to cover all combinations/cases. 

They seem to have been successful, apart from one run that failed in the image preload steps, which is product unrelated! I have re-triggered the run, and asked Copilot to propose a more robust implementation. Let's see where this will lead, but this should not be blocking

MattiaSarti (approved):
> @MattiaSarti I have added ambient tests to cover all combinations/cases. 

Wonderful, thanks!

> They seem to have been successful, apart from one run that failed in the image preload steps, which is product unrelated! I have re-triggered the run, and asked Copilot to propose a more robust implementation. Let's see where this will lead, but this should not be blocking

Rerunning made it green, so no problem!

PR #232, opened 2026-07-21
feat: update MLflow to 3,14

MattiaSarti:
resolves https://github.com/canonical/charmed-kubeflow-solutions/issues/221

note: tested in conjunction with https://github.com/canonical/charmed-kubeflow-uats/pull/287

note: while the changes to `.github/dependencies.yaml` will be undone before merging, the change to `terraform/products/kubeflow/variables.tf` should be kept on `main`, in my opinion


terraform/products/kubeflow/variables.tf
```diff
@@ -6,7 +6,7 @@
 variable "release" {
   type        = string
   description = "Kubeflow release to deploy. Use 'latest' for latest tracks or '1.11' for pinned 1.11 tracks."
-  default     = "1.11"
+  default     = "latest"
```
dariofaccin, line 9: thought: I think using `1.11` as default is advised here, since this product can (and should, hopefully) be consumed by external users and customers. Using `latest` here means deploying a possibly broken deployment, which we do not want to officially support.
I solved this in one of my PRs ([link](https://github.com/canonical/charmed-kubeflow-solutions/pull/218/changes/f5fd06514b6df9af440c7899d6177a52af5f76f7)), where we enforce `latest` in CI only.

terraform/components/mlflow/variables.tf
```diff
@@ -227,7 +227,7 @@ variable "mlflow_server" {
   description = "Configuration for mlflow-server application"
   type = object({
     app_name    = optional(string, "mlflow-server")
-    channel     = optional(string, "2.22/stable")
+    channel     = optional(string, "3.14/edge")
```
dariofaccin: thought: I am not sure we want an `edge` risk to be the default one., given that we can override it at product level.

terraform/products/kubeflow/variables.tf
MattiaSarti: I see your point, but shouldn't users expect that if they deploy the solutions from `main` then we default to `latest` and that `1.11` is the default only as long as they deploy the solutions from `track/1.11`?

terraform/components/mlflow/variables.tf
MattiaSarti: The fact is that this pull request is to switch from MLflow 2.22 to MLflow 3.14, and MLflow 3.14 is only available on `edge` risk so far, so there is no other way to have MLflow 3.14 by default than using `edge` risk here

NohaIhab: you're right @MattiaSarti, however I think we can add a comment to update it to stable once it is released.

terraform/products/kubeflow/variables.tf
NohaIhab: I think for `main` we want to use `1.11/edge`, running `latest/edge` as a solution would be too broken as @dariofaccin said

MattiaSarti: todo: okay, then I could wait for @dariofaccin's PR to be merged to undo this change and rebase to include his workaround and achieve the same overall effect in the CI, but then we would still have the same problem that deploying from `main` with the default `release` would deploy MLflow 2.22 instead of 3.14, so in order to cope with that I'll have to modify the MLflow channel also for the 1.11 release to point to `3.14/edge` - do you find that reasonable, @dariofaccin and @NohaIhab? (I've already updated the source, if you want to take a look)

dariofaccin: @deusebio
I have the feeling that the `release` variable is making things worses on the long run.
I am thinking about dropping it in favor of the following:
* in `main` we only use `latest/<risk>` as channel for all charms (when applicable, there might be exceptions)
* in track branches we pin the track (e.g., `2.22` or `3.14`) and use the `risk` var

Wdyt?

terraform/components/mlflow/variables.tf
dariofaccin: After speaking with @deusebio we agreed that we should maintain `stable` at component level. We will migrate to `3.14/stable` when it will be available.

terraform/products/kubeflow/locals.tf
```diff
@@ -40,7 +40,11 @@ locals {
   resource_dispatcher_channel = var.release == "1.11" ? "2.0/${var.risk}" : "latest/${var.risk}"
 
   # MLflow Component
-  mlflow_channel = var.release == "1.11" ? "2.22/${var.risk}" : "latest/${var.risk}"
+  # NOTE: the risk is forced to `edge` for tracks `3.14` and `latest` as MLflow 3.14 is available
+  # only on `3.14/edge` and `latest/edge` so far - also, mind that changing risk on `latest` would
+  # end up deploying MLflow 2.22
+  # TODO: parametrize the risk as soon as MLflow 3 is promoted, on tracks `3.14` and `latest`
+  mlflow_channel = var.release == "1.11" ? "3.14/edge" : "latest/edge"
```
dariofaccin: issue(blocking): after talking with @deusebio we **do not** want `3.14` on release `1.11`.

dariofaccin (changes requested):

terraform/products/kubeflow/variables.tf
dariofaccin: We agreed to maintain the `release` variable. We will implement a dictionary-like local to set tracks for the various CKF versions.

MattiaSarti:
note: waiting for https://github.com/canonical/charmed-kubeflow-solutions/pull/218 to be merged in order to rebase on that so as to leave the default `release=latest` value that it introduces for CI runs

MattiaSarti:
(closed by mistake while force-pushing)

.github/dependencies.yaml
```diff
@@ -1,5 +1,5 @@
-CKF_BRANCH: "main"
+CKF_BRANCH: "KF-8852/feat/update-mlflow-to-3-14"
 K8S_VERSION: "1.32"
 JUJU_VERSION: "3.6"
-UATS_BRANCH: "main"
+UATS_BRANCH: "KF-8742/feat/update-tests-for-mlflow-3"
```
deusebio: **todo** just dropping a comment to remember to swap this once the UATs PR gets merged

NohaIhab: +1

PR #235, opened 2026-07-22
feat: run `kubeflow-ambient-iam` deployment and tests in CI

NohaIhab:
Closes #228

Adds CI coverage for the **`kubeflow-ambient-iam`** solution (Charmed Kubeflow on the Istio **ambient** mesh + the **Canonical Identity Platform**) on both Canonical K8s and MicroK8s, running the deployment and the ambient + M2M-Identity UATs. Also includes a solution fix surfaced by this testing.

## Summary

- **Deployment test** — applies the `kubeflow-ambient-iam` Terraform root (`istio-system` / `iam` / `iam-core` / `kubeflow` models), configures DNS for the external hostnames, and waits for every model to reach active/idle.
- **UATs** — runs the regular ambient UATs plus the new **M2M Identity** suite (`--include-m2m-tests`, canonical/charmed-kubeflow-uats#281), and raises `github-profiles-automator`'s `sync-period` so it doesn't reconcile mid-run.
- **Workflows** — adds the iam job to the Canonical K8s and MicroK8s validation workflows (larger runner, extended deploy timeout, per-model status/debug capture, iam-only larger controller root-disk on K8s).
- **Fix** — wire only the **UI** ambient gateway to `istio-ingress-config`, this relation is not needed for JWT-only Ingress (confirmed with Service Mesh team)
- **Marked workarounds** — roll out `oauth2-proxy` when it settles idle-but-not-active (canonical/oauth2-proxy-k8s-operator#278); apply the iam Terraform serially (juju/terraform-provider-juju#1308).

## Before merging

* Revert the temporary testing scaffolding marked with `TODO(kf-8863)`: restore the full CI matrices and re-enable the other workflow triggers.


NohaIhab:
@copilot resolve the merge conflicts in this pull request

.github/workflows/on_pull_request.yaml
```diff
@@ -5,7 +5,9 @@ name: On Pull Request
 # * create backport labels if it is against main, only when the PR is opened/reopened
 
 on:
-  pull_request:
+  # TODO(kf-8863): temporarily switched from pull_request to workflow_dispatch so only
+  # deploy-to-k8s.yaml runs on this test PR. Restore `pull_request:` before merge.
+  workflow_dispatch:
```
MattiaSarti: todo: to restore, as you rightly pointed out

terraform/products/iam/main.tf
```diff
@@ -48,7 +48,7 @@ resource "juju_application" "traefik" {
     name     = "traefik-k8s"
     channel  = var.traefik_channel
     revision = var.traefik_revision
-    base     = "ubuntu@26.04"
+    base     = "ubuntu@20.04"
```
MattiaSarti, line 51: question: why is this necessary?

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -26,31 +26,51 @@ on:
...
       max-parallel: 3
       matrix:
-        product: [kubeflow, mlflow, feast, spark]
-        service-mesh-type: [sidecar, ambient]
-        auth-type: [dex]
+        # TODO(kf-8863): temporarily reduced to only the kubeflow-ambient-iam job
+        # to test this solution in CI. Restore the full matrix below before merge.
+        product: [kubeflow]
+        service-mesh-type: [ambient]
+        auth-type: [iam]
```
MattiaSarti: todo: to restore, as you rightly pointed out

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -157,7 +183,7 @@ jobs:
           terraform --version
 
       - name: Deploy
-        timeout-minutes: 45
+        timeout-minutes: ${{ matrix.auth-type == 'iam' && 120 || 45 }}
```
MattiaSarti: thought: does it really take so long that you had to increase the timeout this much? I am concerned about running this in our CI every time, then

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -208,11 +234,27 @@ jobs:
         run: juju status -m kubeflow --relations
         if: always()
 
+      - name: Get juju status for istio and iam models
+        if: always() && matrix.auth-type == 'iam'
```
MattiaSarti, line 223: nitpick: is `always() &&` required?

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -142,7 +156,7 @@ jobs:
           terraform --version
 
       - name: Deploy solution
-        timeout-minutes: 45
+        timeout-minutes: ${{ matrix.auth-type == 'iam' && 120 || 45 }}
```
MattiaSarti: thought: does it really take so long that you had to increase the timeout this much? I am concerned about running this in our CI every time, then

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -26,28 +26,42 @@ on:
...
+    timeout-minutes: 150
 
     strategy:
       fail-fast: false
       max-parallel: 3
       matrix:
-        service-mesh-type: [sidecar, ambient]
-        product: 
+        # TODO(kf-8863): temporarily reduced to only the kubeflow-ambient-iam job
+        # to test this solution in CI. Restore the full matrix below before merge.
```
MattiaSarti: todo: to restore, as you rightly pointed out

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -26,28 +26,42 @@ on:
...
           - policy: privileged
             istio-cni-bin-dir: ""  # meaning Istio CNI disabled
             istio-cni-conf-dir: ""  # meaning Istio CNI disabled
-          - policy: baseline
-            istio-cni-bin-dir: /var/snap/microk8s/current/opt/cni/bin
-            istio-cni-conf-dir: /var/snap/microk8s/current/args/cni-network
-          
+        auth-type: [iam]
+        uats-extra-args: ["--include-ambient-tests --include-m2m-tests"]
+        # --- Full matrix (restore before merge) ---------------------------------
```
MattiaSarti: suggestion (blocking): do not simply restore this commented-out matrix but expand it to also include the new `auth-type` and `uats-extra-args` for `kubeflow` on `ambient` (parametrizing `auth-type` as `dex` for other combinations)

terraform/tests/conftest.py
```diff
@@ -173,11 +173,57 @@ def enable_spark(request) -> list[str]:
     return []
 
 
+@pytest.fixture(scope="module")
+def solution_module_path(request) -> str:
+    """Path to the Terraform root module to apply for the selected auth stack."""
+    if request.config.getoption("--auth-type") == "iam":
+        return "./../tests/kubeflow-ambient-iam"
```
MattiaSarti, line 181: question: I didn't review previous IAM PRs, but how come the Terraform source for the IAM solution lives under `tests`?

terraform/tests/conftest.py
```diff
@@ -173,11 +173,57 @@ def enable_spark(request) -> list[str]:
...
+    if request.config.getoption("--auth-type") == "iam":
+        # The kubeflow-ambient-iam root always deploys ambient + iam and lets
+        # Terraform create the istio-system, iam and iam-core models. Only the
+        # kubeflow model is pre-created by the test and referenced via model_uuid.
+        return (
+            enable_mlflow
+            + enable_feast
+            + istio_k8s_platform
+            + risk
+            + [
```
MattiaSarti, line 215: suggestion (non-blocking): I know `service_mesh_type` and `auth_type` are hardcoded [here](https://github.com/canonical/charmed-kubeflow-solutions/blob/main/terraform/tests/kubeflow-ambient-iam/main.tf#L108-L109), but isn't is worth introducing an assertion for them not to be passed as options with different values from the hardcoded ones, to avoid cases where we think we set them differently and we actually don't?

terraform/tests/conftest.py
```diff
@@ -173,11 +173,57 @@ def enable_spark(request) -> list[str]:
...
+                "-var",
+                "external_m2m_hostname=api.kubeflow.com",
+                "-var",
+                "external_auth_hostname=auth.kubeflow.com",
+                "-var",
+                (
+                    "github_profiles_automator_config={"
+                    'repository="https://github.com/canonical/github-profiles-automator.git",'
+                    '"pmr-yaml-path"="tests/samples/pmr-sample-full.yaml",'
+                    '"git-revision"="main",'
```
MattiaSarti: suggestion (non-blocking): shall we pin a commit hash, for reproducibility?

terraform/tests/conftest.py
```diff
@@ -173,11 +173,57 @@ def enable_spark(request) -> list[str]:
...
+                    # Slow the reconcile so it does not remove the m2m UATs'
+                    # directly-created authorization mid-run.
+                    '"sync-period"="86400"}'
```
MattiaSarti, line 233: praise: well done, for testing

terraform/tests/conftest.py
MattiaSarti, line 215: suggestion (non-blocking): why don't we include the PSS parametrization? It should still work with the IAM integration

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -26,28 +26,42 @@ on:
...
-            uats-env: mlflow-remote
-          - name: feast
-            uats-env: feast-remote
         pod-security-standards:
           - policy: privileged
             istio-cni-bin-dir: ""  # meaning Istio CNI disabled
             istio-cni-conf-dir: ""  # meaning Istio CNI disabled
-          - policy: baseline
-            istio-cni-bin-dir: /var/snap/microk8s/current/opt/cni/bin
-            istio-cni-conf-dir: /var/snap/microk8s/current/args/cni-network
```
MattiaSarti, line 49: suggestion (non-blocking): why don't we include the PSS parametrization? It should still work with the IAM integration (`istio-cni-bin-dir` and `istio-cni-conf-dir` are not used, but `policy` is)

terraform/tests/test_deployment.py
MattiaSarti, line 1: praise: excellent work documenting the workarounds

MattiaSarti, line 132: questions: don't the same assertions apply to IAM mode? Or is the initial login page different, with IAM?

terraform/tests/test_deployment.py
```diff
@@ -95,7 +145,180 @@ def get_public_url(
...
+
+def configure_dns(lightkube_client: lightkube.Client) -> dict[str, str]:
+    """Resolve the ambient-iam hostnames to their LoadBalancer IPs.
+
+    Both the in-cluster DNS (CoreDNS) and the runner host (/etc/hosts) are
+    updated so the IAM charms and cross-model callbacks can resolve the
+    external hostnames and reconcile to active.
+    """
+    host_to_ip = {
+        "ui.kubeflow.com": _wait_for_lb_ip(
```
MattiaSarti: nitpick: can we use shared constants instead of repeating these DNSs? Both here and in `terraform/tests/conftest.py`

terraform/tests/test_deployment.py
```diff
@@ -95,7 +145,180 @@ def get_public_url(
...
+    # CoreDNS does not always pick up the patched Corefile promptly, so force a
+    # rollout to apply the new host entries immediately.
+    subprocess.run(
+        [
+            "kubectl",
+            "-n",
+            "kube-system",
+            "rollout",
+            "restart",
+            "deployment/coredns",
```
MattiaSarti, line 381: question: is the deployment name the same on Canonical K8s and on MicroK8s?

MattiaSarti (changes requested):
Well done with the workarounds!

I have some small concerns:

terraform/tests/conftest.py
deusebio: I believe it is fine to test one specific value of PSS of our choice. We could give it a go with "baseline" since it is the more restrictive one, but the deployment modules here are meant to be a reference architecture that we deploy to show case the integration. it is not meant of being a fully parametrized TF modules to be consumed.

terraform/products/iam/main.tf
NohaIhab: This is the base that gets deployed by the IAM solution, if you follow their [tutorial](https://canonical-identity.readthedocs-hosted.com/tutorial/canonical-identity-platform/) but use `v1.1.1` (the version we vendor here), you get traefik with base `20.04`, when we were deploying `26.04` it was causing problems with the charm in `error` which makes sense because they're not supporting that version with IAM.

.github/workflows/deploy-to-k8s.yaml
NohaIhab: if an `if:` has no status function (success()/failure()/always()/cancelled()), an implicit `success()` is applied, and we want to run this whether it's success, failure, or cancelled. See https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#status-check-functions.

terraform/tests/conftest.py
NohaIhab: see Enrico's comment [here](https://github.com/canonical/charmed-kubeflow-solutions/pull/220#discussion_r3613219867)

terraform/tests/test_deployment.py
NohaIhab: The deployment yes, but the configmap name is different so that's why I have the discovery there. You can see in the CI logs that it's getting restarted for both.

terraform/tests/test_deployment.py
NohaIhab: they're different with IAM because these literals are from dex, I can identify what the contents are and add similar assertions to IAM.

.github/workflows/deploy-to-k8s.yaml
NohaIhab: no not really, this was just experimental to see how long it takes. I updated it to `60`, should be reasonable.

.github/workflows/deploy-to-microk8s.yaml
NohaIhab: no not really, this was just experimental to see how long it takes. I updated it to `60`, should be reasonable.

.github/workflows/deploy-to-microk8s.yaml
MattiaSarti: Agreed not to test IAM with `baseline` PSS in this pull request

terraform/products/kubeflow/variables.tf
```diff
@@ -6,7 +6,7 @@
 variable "release" {
   type        = string
   description = "Kubeflow release to deploy. Use 'latest' for latest tracks or '1.11' for pinned 1.11 tracks."
-  default     = "1.11"
+  default     = "latest"
```
MattiaSarti, line 9: todo: I am completely fine with this, but let's check with @deusebio before merging as he and @dariofaccin preferred to avoid this default at the product level [in my MLflow 3 PR](https://github.com/canonical/charmed-kubeflow-solutions/pull/232) and instead suggested:
- either waiting for [this pull request](https://github.com/canonical/charmed-kubeflow-solutions/pull/218) where `release=latest` is enforced only in the CI to be merged first, to then rebase to that (and that's what I am doing with my PR)
- or cherry-picking [the very commit that does the enforcement](https://github.com/canonical/charmed-kubeflow-solutions/pull/218/changes/f5fd06514b6df9af440c7899d6177a52af5f76f7) from that same pull request

NohaIhab: Hi @MattiaSarti, I already checked with @deusebio that I can do it in this PR. The other PRs can rebase on main once I merge, is that ok with you?

MattiaSarti: Absolutely! Go ahead and I'll rebase on yours, then

NohaIhab:
I'm seeing some flakiness in the CI so I will pause on merging this while I try to debug it in #245 

Update: I discovered an issue in Hydra, and filed it with details in https://github.com/canonical/hydra-operator/issues/591. Implemented a workaround for it in https://github.com/canonical/charmed-kubeflow-solutions/pull/235/commits/70cd13aeba7f20e688e447768577b961a284bb95.

PR #241, opened 2026-07-24
feat: migrate to UATs refactoring branch

dariofaccin:


.github/dependencies.yaml
```diff
@@ -1,5 +1,5 @@
 CKF_BRANCH: "main"
 K8S_VERSION: "1.32"
 JUJU_VERSION: "3.6"
-UATS_BRANCH: "main"
+UATS_BRANCH: "feat/refactor-uats"
```
MattiaSarti: todo: let's restore once merged

.github/workflows/deploy-to-microk8s.yaml
```diff
@@ -261,55 +263,20 @@ jobs:
...
-
       - name: Get juju status
-        run: juju status
-        if: failure() || cancelled()
-
-      - name: Get juju debug logs
-        run: juju debug-log --replay --no-tail
-        if: failure() || cancelled()
+        run: juju status -m kubeflow --relations
+        if: always()
```
MattiaSarti: nitpick: `if: always()` alone shouldn't be necessary

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -207,23 +221,38 @@ jobs:
...
+          LOG_DIR="${{ github.workspace }}/test-kubeflow-pod-logs"
+          mkdir -p "$LOG_DIR"
+          for pod in $(kubectl get pods -n test-kubeflow -o jsonpath='{.items[*].metadata.name}'); do
+            kubectl logs -n test-kubeflow "$pod" --all-containers --prefix > "$LOG_DIR/${pod}.log" 2>&1 || true
+            kubectl logs -n test-kubeflow "$pod" --all-containers --prefix --previous > "$LOG_DIR/${pod}-previous.log" 2>&1 || true
+          done
+
+      - name: Upload test-kubeflow pod logs
+        if: failure() && steps.collect-pod-logs.outcome == 'success'
+        uses: actions/upload-artifact@v4
```
MattiaSarti: question: why do we do this only for Canonical K8s?

.github/workflows/deploy-to-microk8s.yaml
dariofaccin: If a previous step fails, without `if: always()` the subsequent steps are not executed.

.github/workflows/deploy-to-k8s.yaml
dariofaccin: Typo.

MattiaSarti (approved):

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -33,30 +33,15 @@ jobs:
       fail-fast: false
       max-parallel: 3
       matrix:
-        product: [kubeflow, mlflow, feast, spark]
-        service-mesh-type: [sidecar, ambient]
-        auth-type: [dex]
+        product: [kubeflow]
+        service-mesh-type: [ambient]
+        auth-type: [iam]
```
deusebio: **question** why are we reducing the testing matrix? is this just for testing?

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -207,23 +206,59 @@ jobs:
...
           tox -c ./terraform/tests -e test-deployment -- -vv -s $RISK_ARG $SERVICE_MESH_TYPE_ARG $AUTH_TYPE_ARG $EXTRA_COMPONENTS_ARG
 
-      - name: Clone UATs repository
+      - name: Run UATs
+        id: run-uats
         run: |
-          git clone https://github.com/canonical/charmed-kubeflow-uats.git ${{ github.workspace }}/charmed-kubeflow-uats
           cd ${{ github.workspace }}/charmed-kubeflow-uats
-          git checkout ${{ env.UATS_BRANCH }}
+          tox -e ${{ matrix.uats-env }} -- ${{ matrix.uats-extra-args }} --bundle="" --keep-artifacts
```
deusebio: **todo** remove the `--bundle`. I believe it has also been working without

deusebio (commented):
A couple of points, but I suppose these are around just for testing and not to be merge in the main branch. Is this correct? If so, I'd generally suggest when opening the PR to just put a comment in the code such that this does not fall through the cracks and we make sure these are indeed addressed before merging.

.github/workflows/deploy-to-k8s.yaml
```diff
@@ -207,44 +206,103 @@ jobs:
...
-          git checkout ${{ env.UATS_BRANCH }}
+          if [[ "${{ matrix.auth-type }}" == "iam" ]]; then
+            poetry install --no-root
+            poetry run playwright install --with-deps chromium
+          fi
+          tox -e ${{ matrix.uats-env }} -- ${{ matrix.uats-extra-args }} --bundle="" --keep-artifacts --filter "m2m"
 
-      - name: Preload Images into Containerd Cache
-        uses: ./.github/actions/preload-images
+      - name: Collect user namespace pods logs
```
mvlassis: thought: We should probably write a reusable action for these collect & upload log actions to avoid duplicate code.

.github/workflows/deploy-to-k8s.yaml
dariofaccin: Only testing. Now we figured out the issues with IAM testing, and will revert once everything is stable enough.

.github/workflows/deploy-to-k8s.yaml
dariofaccin: I created a new action to fetch and upload k8s logs. It always collect pods logs from the input namespaces and possibly namespace prefixes (to gather all the `test-m2m` namespaces too). It also accepts a list of resource kinds to dump.

